PR #6, opened 2018-09-08
cloud-init/daily_deb.sh: retry failures- add-apt-repostory or apt update

blackboxsw:
Add a retry_on_fail function which will retry X times with an incremental
5 second backoff per retry. On failure, emit failure message.


blackboxsw:
While retries might be helpful in some occasions, probably not going to accept this branch as this particular problem isn't timing related. It affects 100% of cosmic containers which end up with no network configured.
https://bugs.launchpad.net/cloud-images/+bug/1791691

PR #9, opened 2019-01-29
Fix the Copr build and test scripts

paride:
Disable the epel-6 build and enable the `baseurl` repository when using an http proxy.


paride:
Tested on torkoal.

PR #10, opened 2019-05-22
Enable boot speed measurements on devices

paride:


boot-speed/testflinger/bootspeed.sh
```diff
@@ -0,0 +1,102 @@
...
+#          hostname ; \
+#          arch ; \
+#          uname -a ; \
+#          id ; \
+#          w ; \
+#          pwd ; \
+#          ip addr ; \
+#          free -m ; \
+#          mount ; \
+#          df -h" > system-info 2>&1
```
powersj: can we clean this up above

boot-speed/testflinger/measure-device.sh
```diff
@@ -0,0 +1,138 @@
...
+
+image_url=$(grep "url:" "$yaml_head" | awk '{ print $2 }')
+image_dirname=$(echo $image_url | sed 's|\(.*/\)\(.*\)|\1|')
+image_basename=$(echo $image_url | sed 's|\(.*/\)\(.*\)|\2|')
+image_serial=$(curl -s --noproxy ubuntu.com $image_dirname/.publish_info | grep $image_basename | awk '{ print $2 }') || true
+
+regexp="^[0-9]{8}(\.[0-9]{1,2})?$"
+if [[ ! "$image_serial" =~ $regexp ]]; then
+    echo "WARNING: Image serial not found! Falling back using today's date."
+    image_serial=$yyyymmdd
```
powersj: This doesn't seem right. When do we not find the image serial? Shouldn't we simply fail?

boot-speed/testflinger/measure-device.sh
```diff
@@ -0,0 +1,138 @@
...
+mkdir -v $datadir
+
+image_url=$(grep "url:" "$yaml_head" | awk '{ print $2 }')
+image_dirname=$(echo $image_url | sed 's|\(.*/\)\(.*\)|\1|')
+image_basename=$(echo $image_url | sed 's|\(.*/\)\(.*\)|\2|')
+image_serial=$(curl -s --noproxy ubuntu.com $image_dirname/.publish_info | grep $image_basename | awk '{ print $2 }') || true
+
+regexp="^[0-9]{8}(\.[0-9]{1,2})?$"
+if [[ ! "$image_serial" =~ $regexp ]]; then
+    echo "WARNING: Image serial not found! Falling back using today's date."
```
paride: Workaround to be removed after the .publish_info files will be in place for Core 16 and Core 18.

boot-speed/testflinger/measure-device.sh
```diff
@@ -0,0 +1,138 @@
...
+cat "$yaml_head" "$yaml_tail" > "$yaml_full"
+
+#testflinger-cli submit "$yaml_full" | tee testflinger-submit-output
+
+#if ! grep -q "^job_id:" testflinger-submit-output; then
+#    echo "Failed to submit job."
+#    exit 1
+#fi
+
+#job_id=$(awk '/^job_id:/{ print $2 }' testflinger-submit-output)
```
powersj: again can we remove commented out code

boot-speed/testflinger/test_cmds.sh
```diff
@@ -0,0 +1,79 @@
...
+cat /etc/os-release
+uname -a
+id
+pwd
+ls -la
+
+# Do not auto-update snaps on the provisioned device for now; we'll
+# update them manually later. If changing this remember that Core
+# may auto-reboot after a snap refresh.
+ssh ubuntu@$DEVICE_IP 'sudo snap set system refresh.hold="$(date --date=tomorrow +%Y-%m-%dT%H:%M:%S%:z)"'
```
powersj: Is there documentation about the auto-reboot after a snap refresh? This does not seem right. Have you talked to the snapd team?

Part of this testing is to gather the snapd refresh time. By putting a hold on the system refresh are you not preventing us from capturing that data?

boot-speed/testflinger/test_cmds.sh
```diff
@@ -0,0 +1,79 @@
...
+
+mkdir artifacts
+cd artifacts
+date --utc --rfc-3339=ns > job-start-timestamp
+
+curl -Ss -O https://raw.githubusercontent.com/paride/server-test-scripts/bootspeed/boot-speed/testflinger/bootspeed.sh
```
powersj: obviously this URL will need to be changed ;)

boot-speed/testflinger/test_data.yaml
```diff
@@ -0,0 +1,19 @@
...
+        # For historical reasons testflinger calls format() on the test_cmds
+        # string to fill in certain values if desided. That means it tries to
+        # interpret anything between curly braces as a key, so literal curly
+        # braces must be escaped by doubling them up.
+        #
+        # To avoid worrying about escaping and for easiear testing/debugging
+        # we keep the string very simple and just pull and run an external
+        # script.
+
+        wget https://raw.githubusercontent.com/paride/server-test-scripts/bootspeed/boot-speed/testflinger/test_cmds.sh
```
powersj: again update the URL

boot-speed/testflinger/measure-device.sh
paride: The publish_info file is not there yet for Core16 and Core18. The file is getting generated only for new builds. Lukasz (sil2100) may generate the missing files manually, but I may have to ping again.

The workaround is there to get successful runs while we don't have a publish_info file for all the images.

boot-speed/testflinger/test_cmds.sh
paride: I didn't read about Core auto rebooting when the core or kernel snaps are refreshed in the official documentation but the thing is well known, and I guess it's written somewhere.

What I'm doing it: measure the boot time before refreshing the snaps, then refresh the snaps (allowing the system to reboot as many times as it wishes), and then measure the boot speed with the updated snaps. The actual refresh time is not recorded. This is different from the time it takes for snapd.seeded.service to complete.

I'm holding the snaps to prevent the system from rebooting while the measurement is being taken.

Measuring the snapd refresh time is not easy. Do we count in the (up to 2) reboots it takes to refresh the snaps? What about the time it takes to download the snaps?

boot-speed/testflinger/measure-device.sh
powersj: What is the ETA for getting those files? I don't want us to start failing into this block and incorrectly reporting the image_serial date. This seems to be an assumption I would like to avoid.

boot-speed/testflinger/test_cmds.sh
powersj: The main point of capturing device boot time is to capture the refresh time as a part of the initial boot. Therefore, we need to have a discussion with the snapd team.

boot-speed/testflinger/measure-device.sh
paride: Anyway yes, it should definitely error when the serial can't be found. Let me ping sil2100 to have those .publish_info files uploaded.

PR #14, opened 2019-05-29
bootspeed: better error and exception handling

paride:
Changes:
 - move the dirty device check in test_cmds
 - always wait and retry when a snapd change status is "Error"
 - be extra sure to error if bootspeed.sh fails
 - cancel the job if testflinger timeouts


paride:
Coming up with a PR which includes the clouds measurement scripts

PR #16, opened 2019-05-30
bootspeed: better handling of errors; cloud measurements

paride:


paride:
Up for review, probably not fully ready for merging. Relevant for the cloud measurements is only the last commit (c6d5f98). One ugly point is:

https://github.com/CanonicalLtd/server-test-scripts/pull/16/files#diff-f513c6a2108c1394a4af4dc6187faad7R163

where I hardcode the SubnetId to be used; it has to be made a parameter.

boot-speed/clouds/measure-cloud.py
```diff
@@ -0,0 +1,242 @@
...
+    except Exception as e:
+        print('An creating SFTP client: %s: %s' % (e.__class__, e))
+        if sftp is not None:
+            sftp.close()
+        if transport is not None:
+            transport.close()
+        pass
+
+
+def cloudid2serial(id):
```
powersj: we should consider adding this to pycloudlib

boot-speed/clouds/measure-cloud.py
```diff
@@ -0,0 +1,242 @@
...
+        mirror_url='https://cloud-images.ubuntu.com/daily',
+        keyring_path='/usr/share/keyrings/ubuntu-cloudimage-keyring.gpg'
+    )
+    filters = ['id=%s' % id]
+    res = stream.query(filters)
+    id = res[0]['id']
+    return id
+
+
+def measure_instance(instance, datadir, reboots=1):
```
powersj: the first part of this function is reused code from devices.

Can we create one function that has different calls depending on the type?

Although, only devices will auto-reboot, so cloud and lxd/kvm should not, so maybe we can remove all of that code as well to make this simpler. Thoughts?

boot-speed/clouds/measure-cloud.py
```diff
@@ -0,0 +1,242 @@
...
+        sftpclient.close()
+
+    for tarball in glob.glob('boot_*.tar.gz'):
+        with tarfile.open(tarball, "r:gz") as tar:
+            tar.extractall(path=str(datadir))
+        os.unlink(tarball)
+
+
+def measure_ec2(
+        release, datadir, *, inst_type="t2.micro", region="us-east-1",
```
powersj: \* for?

paride: To force the parameters following the * (`inst_type` and `region` in this case) to be keyword-only, and not positional. While developing it happened to me to use them in the wrong order, so I decided to force the usage the `param=value` syntax when calling the function. That's all; the * is not necessary, it just makes the syntax a bit more strict.

boot-speed/clouds/measure-cloud.py
paride: +1, it belongs there, but I'd like to see the job working first, to be sure there are no bigger issues around. I'll have to start working on pycloudlib to add the missing clouds at some point, that will probably be a good moment to add this function too.

boot-speed/clouds/measure-cloud.py
paride: I can see your point, but I'm not sure it's worth it. The two cases (testflinger devices vs. clouds) are indeed similar but different enough to require to be handled differently, at least in part. The auto-reboot thing is one example. The reused commands are just a few, while the common part is in `bootspeed.sh`, which is exactly the same for devices, clouds, or any other system we may want to measure.

A couple of commands could be moved there, e.g. the boot_id check. The `rm -rf artifacts` is redundant (bootspeed.sh already does it), and it's there just for extra safety, as we want to be really sure not to retrieve and old measurement. What remains is very little.

PR #17, opened 2019-06-05
bootspeed: debugging printouts if systemd timeout is reached

paride:


panlinux:
I would also suggest a process listing dump, like `ps fauxw`

paride:
Thanks @panlinux

boot-speed/bootspeed.sh
```diff
@@ -52,6 +52,8 @@ if timeout 20m sh -c "until systemd-analyze time; do sleep 20; done"; then
     systemd-analyze critical-chain > systemd-analyze_critical-chain
     systemd-analyze plot > systemd-analyze_plot.svg
 else
```
OddBloke: The output of `systemd-analyze time` in the failure case includes "Hint: Use 'systemctl list-jobs' to see active jobs", is it worth including the output of `systemctl list-jobs` here? (`blame` may capture the same info, I'm not sure.)

paride: Yeah I'm refactoring a bit that part of the script and added a `list-jobs`. Thanks!

PR #19, opened 2019-06-11
bootspeed: better support for EC2 measurements

paride:
* refactor measure-cloud.py
* add an EC2 cleanup script

### DIFF
--- a/boot-speed/clouds/ec2_cleanup.py
+++ b/boot-speed/clouds/ec2_cleanup.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+"""Clean up long-running (= stale) bootspeed-* EC2 instances.
+
+Copyright 2018-2019 Canonical Ltd.
+Joshua Powers <josh.powers@canonical.com>
+Paride Legovini <paride.legovini@canonical.com>
+"""
+import boto3
+import datetime as dt
+import re
+
+
+def clean_ec2():
+    """Clean up all running EC2 instances tagged 'bootspeed-*'."""
+
+    # Seconds since epoch; will be compared with the timestamp we put
+    # in the instance tag to determine if instances are stale.
+    now = int(dt.datetime.utcnow().timestamp())
+
+    resource = boto3.resource('ec2')
+    inst_tag = [{'Name': 'tag:Name', 'Values': ['bootspeed-*']}]
+    stale_instances = []
+    for vpc in list(resource.vpcs.all()):
+        print('# cleaning up vpc %s\n' % vpc.id)
+        for instance in vpc.instances.filter(Filters=inst_tag):
+            for tag in instance.tags:
+                if tag['Key'] == 'Name' and re.match(
+                        "^bootspeed-[0-9]+$", tag['Value']):
+                    print('Found bootspeed instance %s' % instance.id)
+                    timestamp = int(tag['Value'].split("-")[1])
+                    tdelta = now - timestamp
+                    print('Instance started %s seconds ago' % tdelta)
+                    if tdelta > 3*3600:
+                        print("Terminating stale instance.")
+                        stale_instances.append(instance)
+                        instance.terminate()
+                    print()
+
+    for instance in stale_instances:
+        print("Waiting for %s to be terminated." % instance.id)
+        instance.wait_until_terminated()
+
+    print("Done")
+
+if __name__ == '__main__':
+    clean_ec2()
--- a/boot-speed/clouds/measure-cloud.py
+++ b/boot-speed/clouds/measure-cloud.py
@@ -15,12 +15,85 @@ import tarfile
 import os
 import paramiko
 import pycloudlib
+import shutil
 import sys
 
 from pathlib import Path
 
 
 known_clouds = ['ec2', 'gce']
+
+
+class EC2Instspec:
+    def __init__(self, *, release, inst_type, region, ec2_subnetid, ec2_sgid):
+        # Defaults. They can't be set as keyword argument defaults because
+        # we're always passing all the arguments to __init__, even if they
+        # are None. And they can't be set as the argparse default values,
+        # as different coulds need different defaults.
+        self.region = "us-east-1"
+        self.inst_type = "t2.micro"
+        self.subnetid = ""
+        self.sgid = []
+
+        # User-specified settings
+        self.release = release
+
+        if inst_type:
+            self.inst_type = inst_type
+        if region:
+            self.region = region
+        if ec2_subnetid:
+            self.subnetid = ec2_subnetid
+        if ec2_sgid:
+            self.sgid = [ec2_sgid]
+
+    def measure(self, datadir, instances=1, reboots=1):
+        """
+        Measure Amazon AWS EC2.
+        Returns the measurement metadata as a dictionary
+        """
+        print('Perforforming measurement on Amazon EC2')
+
+        ec2 = pycloudlib.EC2(tag='bootspeed', region=self.region)
+
+        if self.inst_type.split('.')[0] == 'a1':
+            daily = ec2.daily_image(release=self.release, arch='arm64')
+        else:
+            daily = ec2.daily_image(release=self.release)
+
+        print("Daily image for", self.release, "is", daily)
+
+        metadata = gen_metadata(
+            "ec2", self.region, self.inst_type, self.release, daily)
+
+        for ninstance in range(instances):
+            instance_data = Path(datadir, "instance_" + str(ninstance))
+            instance_data.mkdir()
+
+            # This tag name will be inherited by the launched instance.
+            # We want it to be unique and to contain an easily parsable
+            # timestamp (UTC seconds since epoch), which we will use to
+            # detemine if an instance is stale and should be terminated.
+            tag = "bootspeed-" + str(int(dt.datetime.utcnow().timestamp()))
+            ec2.tag = tag
+
+            print("Launching instance", ninstance+1,
+                  "of", instances, "tag:", ec2.tag)
+            instance = ec2.launch(
+                daily,
+                instance_type=self.inst_type,
+                SubnetId=self.subnetid,
+                SecurityGroupIds=self.sgid
+            )
+            print("Instance launched.")
+
+            try:
+                measure_instance(instance, instance_data, reboots)
+            finally:
+                print("Deleting the instance.")
+                instance.delete(wait=False)
+
+        return metadata
 
 
 def create_sftp_client(host, user, port=22, password=None, keyfilepath=None):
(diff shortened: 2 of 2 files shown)


boot-speed/clouds/measure-cloud.py
```diff
@@ -200,43 +215,65 @@ def gen_datadirname(metadata):
...
     PARSER.add_argument('--reboots', help='Number of reboots',
                         default=1, type=int)
     PARSER.add_argument('--instances', help='Number of instances',
                         default=1, type=int)
+    PARSER.add_argument('--ec2-subnetid', help='AWS EC2 SubnetId')
+    PARSER.add_argument('--ec2-sgid', help='AWS EC2 SecurityGroupId')
+    PARSER.add_argument('--region', help='Cloud region')
     ARGS = PARSER.parse_args()
-    main(ARGS.cloud, ARGS.release, ARGS.instances, ARGS.reboots)
+    return(ARGS)
```
OddBloke: No need for the parentheses here

boot-speed/clouds/ec2_cleanup.py
```diff
@@ -0,0 +1,36 @@
...
+    inst_tag = [{'Name': 'tag:Name', 'Values': ['bootspeed-*']}]
+    for vpc in list(resource.vpcs.all()):
+        print('# cleaning up vpc %s\n' % vpc.id)
+        for instance in vpc.instances.filter(Filters=inst_tag):
+            for tag in instance.tags:
+                if tag['Key'] == 'Name' and re.match(
+                        "^bootspeed-[0-9]+-", tag['Value']):
+                    print('Found bootspeed instance %s' % instance.id)
+                    timestamp = int(tag['Value'].split("-")[1])
+                    tdelta = int(dt.datetime.utcnow().timestamp()) - timestamp
```
OddBloke: I would suggest calculating "now" once so that the script has a consistent baseline that it's operating against; my past experience suggests that this makes behaviour easier to understand.

boot-speed/clouds/ec2_cleanup.py
```diff
@@ -0,0 +1,36 @@
...
+                    print('Instance started %s seconds ago' % tdelta)
+                    if tdelta > 4*3600:
+                        print("Deleting stale instance.")
+                        instance.terminate()
+                        instance.wait_until_terminated()
```
OddBloke: This script is likely to run much faster if you separate the initiate-termination and wait-for-termination in to separate phases (because then all the terminations are happening in parallel). (I think this should be as simple as appending `instance` to a list and then just iterating over that list calling `wait_until_terminated`?)

boot-speed/clouds/measure-cloud.py
```diff
@@ -15,13 +15,66 @@
...
 
 known_clouds = ['ec2', 'gce']
 
+class EC2Instspec():
+    def __init__(self, release):
+        self.release = release
+
+    # Defaults
+    region = "us-east-1"
+    inst_type = "t2.micro"
```
OddBloke: What's the rationale for using t2.micro instances?

(I know this is just a refactor, so I don't want to block on this question.)

boot-speed/clouds/ec2_cleanup.py
```diff
@@ -0,0 +1,36 @@
...
+Joshua Powers <josh.powers@canonical.com>
+Paride Legovini <paride.legovini@canonical.com>
+"""
+import boto3
+import datetime as dt
+import re
+
+
+def clean_ec2():
+    """Clean up all running EC2 instances tagged 'bootspeed-*'."""
```
OddBloke: This won't actually clean up _all_ instances.

boot-speed/clouds/measure-cloud.py
```diff
@@ -200,43 +215,65 @@ def gen_datadirname(metadata):
...
+    args = parse_args()
+
+    if args.cloud not in known_clouds:
         print('Unknown cloud provider:', cloud)
         sys.exit(1)
 
+    if args.cloud == 'ec2':
+        instspec = EC2Instspec(args.release)
+        if args.inst_type:
+            instspec.inst_type = args.inst_type
```
OddBloke: Should these be parameters on `__init__` instead? I can't see anywhere that we use them as class attributes.

boot-speed/clouds/measure-cloud.py
```diff
@@ -200,43 +215,65 @@ def gen_datadirname(metadata):
...
-        raise NotImplementedError
+    metadata = instspec.measure(tmp_datadir, args.instances, args.reboots)
 
     # str() needed for compatibility with Python <= 3.5
     with open(str(Path(tmp_datadir, "metadata.json")), 'w') as mdfile:
         json.dump(metadata, mdfile)
 
     datadir = gen_datadirname(metadata)
-    with tarfile.open(str(datadir.with_suffix(".tar.gz")), "w:gz") as tar:
+    with tarfile.open((datadir + ".tar.gz"), "w:gz") as tar:
```
OddBloke: `datadir` here is presumably not a directory (else we're creating a very strangely named tarball!); could we make the name a little more descriptive?

boot-speed/clouds/measure-cloud.py
```diff
@@ -200,43 +215,65 @@ def gen_datadirname(metadata):
...
         tar.add("data", arcname=str(datadir))
```
OddBloke: Should `"data"` instead be referencing `tmp_datadir` in some way?

boot-speed/clouds/measure-cloud.py
```diff
@@ -200,43 +215,65 @@ def gen_datadirname(metadata):
...
 
+    shutil.rmtree("data")
```
OddBloke: Same question: should `"data"` instead be referencing tmp_datadir in some way?

boot-speed/clouds/measure-cloud.py
```diff
@@ -15,13 +15,66 @@
...
 import pycloudlib
+import shutil
 import sys
 
 from pathlib import Path
 
 
 known_clouds = ['ec2', 'gce']
 
+class EC2Instspec():
```
OddBloke: No need for these parentheses.

boot-speed/clouds/measure-cloud.py
```diff
@@ -15,13 +15,66 @@
...
+    sgid = []
+
+    def measure(self, datadir, instances=1, reboots=1):
+        """
+        Measure Amazon AWS EC2.
+        Returns the measurement metadata as a dictionary
+        """
+        print('Perforforming measurement on Amazon EC2')
+
+        logging.basicConfig(level=logging.DEBUG)
```
OddBloke: Is this so that we see the logs from the libraries we're using? It doesn't look like we're logging anywhere.

(Doing this in `measure` also seems quite far down the stack; should we be doing it in `main()`?)

boot-speed/clouds/measure-cloud.py
```diff
@@ -15,13 +15,66 @@
...
+                SubnetId=self.subnetid,
+                SecurityGroupIds=self.sgid
+            )
+            print("Instance launched.")
+
+            try:
+                measure_instance(instance, instance_data, reboots)
+            finally:
+                print("Deleting the instance.")
+                instance.delete(wait=False)
```
OddBloke: As we don't wait for `delete`, do we run in to any problems with reusing the same name/tag when running more than a single instance?

boot-speed/clouds/ec2_cleanup.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+"""Clean up all running bootspeed-* EC2 instances.
+
+Copyright 2018-2019 Canonical Ltd.
```
OddBloke: I assume this is 2018-2019 because we're importing this script in to this repo from elsewhere?

boot-speed/clouds/measure-cloud.py
paride: It just a default, chosen for two reasons:
* It's pycloudlib's default too
* It's *not* an instance type we are interested in measuring for the boot-speed KPI, the actual instance types we want are specified by the Jenkins job using the command line parameter. If a job ends up measuring t2.micro it means that I forgot to specify the instance type, or more in general that something went wrong :)

boot-speed/clouds/ec2_cleanup.py
paride: Yes, I took

https://github.com/CanonicalLtd/server-test-scripts/blob/master/cloud-init/ec2_cleanup.py

in this same repo and adapted it.

boot-speed/clouds/measure-cloud.py
paride: I'm mostly interested in logging what pycloudlib is doing, but I agree it makes more sense to have it in `main()`, so I moved it there. It logs to stderr (and thus to the Jenkins console log).

boot-speed/clouds/measure-cloud.py
paride: I now make sure that every instance has a different tag. I think we are going to run only one instance per job, but you are right: if running more than one this could have caused some headaches. Thanks!

boot-speed/clouds/measure-cloud.py
```diff
@@ -23,6 +24,75 @@
 known_clouds = ['ec2', 'gce']
 
 
+class EC2Instspec:
+    def __init__(self, *, release, inst_type, region, ec2_subnetid, ec2_sgid):
+        # Defaults
+        self.region = "us-east-1"
```
OddBloke: These could just be the keyword argument defaults, I think?

(Except for `sgid`, because using mutable objects as defaults is an anti-pattern: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html)

paride: I did it like this because the keyword arguments are all getting passed to `__init__`, even if they are `None`, and thus the keyword defaults are always getting overwritten. And I can't set the defaults as the `argparse` arguments default values, as I need different values for different clouds.

But I may be missing the right way to do this, and I'd be happy to, as I don't really like how this is currently implemented, so suggestions are welcome!

OddBloke: Ah, hm, that is tricky. You _could_ construct a kwarg dict dynamically in the call site:

```py
kwargs = {}
if args.release:
    kwargs['release'] = args.release
...
EC2Instspec(**kwargs)
```
but that's really just moving the args handling from one place to another, which doesn't gain us a huge amount.

I think we can just add a comment with what you explained above and move on. :-)

OddBloke:
Thanks! WIll leave to you to merge.

PR #20, opened 2019-06-13
bootspeed: make measure-device.sh work independently of the CWD

paride:
The change will make things easier to handle on the Jenkins side.


paride:
I created the PR with hub(1) and it picked up and old PR title for some reason. It's fixed now.

PR #24, opened 2019-06-20
bootspeed: better support for multiple runs

paride:
* use file timestamps up to the second
 * sleep between reboots

### DIFF
--- a/boot-speed/clouds/measure-cloud.py
+++ b/boot-speed/clouds/measure-cloud.py
@@ -22,6 +22,7 @@ from pathlib import Path
 
 
 known_clouds = ['ec2', 'gce']
+job_timestamp = dt.datetime.utcnow()
 
 
 class EC2Instspec:
@@ -217,11 +218,11 @@ def gen_metadata(
         *, cloud, region, availability_zone='', inst_type, release, cloudid,
         serial):
     """ Returns the instance metadata as a dictionary """
-    yyyymmdd = dt.datetime.utcnow().strftime('%Y%m%d')
-    isodate = dt.datetime.utcnow().isoformat()
+    date = job_timestamp.strftime('%Y%m%d%H%M%S')
+    isodate = job_timestamp.isoformat()
 
     metadata = {}
-    metadata['date'] = yyyymmdd
+    metadata['date'] = date
     metadata['date-rfc3339'] = isodate
     metadata['type'] = "cloud"
     metadata['instance'] = {
@@ -239,13 +240,13 @@ def gen_metadata(
 
 def gen_archivename(metadata):
     """ Generate a standardized measurement directory (and tarball) name """
-    yyyymmdd = metadata['date']
+    date = metadata['date']
     cloud = metadata['instance']['cloud']
     inst_type = metadata['instance']['instance_type']
     release = metadata['instance']['release']
 
-    datadir = cloud + "-" + inst_type + "-" + release + "_" + yyyymmdd
-    return datadir
+    arcname = cloud + "-" + inst_type + "-" + release + "_" + date
+    return arcname
 
 
 def main():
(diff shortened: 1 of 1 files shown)


paride:
Follows @OddBloke's suggestion of setting a "now" timestamp variable once for all in a script to improve readability and consistency.

boot-speed/clouds/measure-cloud.py
```diff
@@ -190,6 +192,7 @@ def measure_instance(instance, datadir, reboots=1):
     for nboot in range(1, reboots+1):
         bootdir = "boot_" + str(nboot)
         instance.restart()
+        time.sleep(60)
```
powersj: what is this for? doesn't restart have a wait flag?

paride: It doesn't, but it probably should:

https://git.launchpad.net/pycloudlib/tree/pycloudlib/ec2/instance.py#n109

Let's do like this if you agree: I'll add a `wait` flag to pycloudlib and update this PR to use it.

powersj: ha! I remember why:

```
# This is a terrible hack, however it is not obviously clear
        # how to wait for a system to start to restart.
        time.sleep(30)
```

I'd like to avoid generically waiting 30 + 60 seconds and solve this in pycloudlib.

OddBloke: This is how we solved this in cloud-test-framework:

```py
def _count_reboots_for_instance(instance):
    return len(
        instance.interactions.run_command('last reboot -F').splitlines())


class RebootInstanceTask(Task):

    def execute(self, test_metadata, instance):
        LOGGER.info('Rebooting instance "{}"...'.format(instance.instance_id))
        pre_reboot_count = _count_reboots_for_instance(instance)
        instance.interactions.trigger_reboot()
        down = False
        while not down:
            try:
                if _count_reboots_for_instance(instance) > pre_reboot_count:
                    break
            except (NetworkError, socket.error, EOFError):
                down = True
            time.sleep(1)
        instance.interactions.purge_ssh_connection()
        LOGGER.info('Instance down; proceeding.')
```

So, basically, determine the number of times the system has rebooted thus far before triggering the reboot. Then reboot and wait until either we detect that the number of reboots has increased (which means it completed the reboot between us attempting to determine if it was down), or if we detect that we can no longer connect to the instance we assume that it's in the process of rebooting.

(In c-t-f, tasks always wait for SSH to be available and cloud-init to complete, so just getting to "down" is sufficient there.)

paride: boto3 doesn't help:

> Requests a reboot of the specified instances. This operation is asynchronous; it only queues a request to reboot the specified instances. 
> 
> If an instance does not cleanly shut down within four minutes, Amazon EC2 performs a hard reboot.

One easy solution here is bumping the sleep time to 60s in pycloudlib. I observed 30s not being enough sometimes, and the default should be safe.

Another solution is adding a `sleep=30` parameter to the `restart()` method, to increase (or decrease) the sleep time without double-sleeping. *Looks* clean but it's a bit ad-hoc for EC2, as we may have better solutions for other clouds, and we want to keep the high-level pycloydlib interface uniform.

paride: I wrote the previous comment before reading @OddBloke's. Let's see what I can come up with.

paride: I dropped the commit adding the sleep and implemented a wait mechanism in pycloudlib:

https://code.launchpad.net/~legovini/pycloudlib/+git/pycloudlib/+merge/369113

paride: An improvement on yesterday's MP:

https://code.launchpad.net/~legovini/pycloudlib/+git/pycloudlib/+merge/369158

It handles the case where, in @OddBloke's code, this fails:
```
pre_reboot_count = _count_reboots_for_instance(instance)
```

paride:
I tested a couple of i3.metal jobs with this changes and the pycloudlib MP merged in and they succeeded.

PR #25, opened 2019-06-24
bootspeed devices: only retry `snap refresh` if Error is in the last change

paride:


paride:
Failure mode of the previous version: http://paste.ubuntu.com/p/MHT8YmNcvZ/

PR #29, opened 2019-07-29
ec2_cleanup: improve the script logic

paride:
- MAX_EC2_INST_AGE: cast to int
 - Use a set for stale_instances
 - break the inner for loop after the first 'Name' tag matching
   'bootspeed-*' is found
 - Retry wait_until_terminated() once if it fails
 - Nicer script output


paride:
tested locally, works fine

PR #31, opened 2019-08-21
measure-cloud: support the release metanames

paride:
Support the Ubuntu release metanames: lts, stable, devel, latest.
They are resolved to actual release names using ubuntu-distro-info(1).

### DIFF
--- a/boot-speed/clouds/measure-cloud.py
+++ b/boot-speed/clouds/measure-cloud.py
@@ -8,6 +8,7 @@ Paride Legovini <paride.legovini@canonical.com>
 
 import argparse
 import datetime as dt
+import distro_info
 import glob
 import json
 import logging
@@ -24,6 +25,7 @@ from pathlib import Path
 
 
 known_clouds = ['kvm', 'lxd', 'ec2', 'gce']
+distro_metanames = ('lts', 'stable', 'latest', 'devel')
 job_timestamp = dt.datetime.utcnow()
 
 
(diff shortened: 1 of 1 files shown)


boot-speed/clouds/measure-cloud.py
```diff
@@ -323,6 +331,10 @@ def gen_archivename(metadata):
     arcname = cloud + "-" + inst_type + "-" + release + "_" + date
     return arcname
 
+def metaname2release(metaname):
```
powersj: Have you looked at python3-distro-info?

```
import distro_info
    try:
        devel = distro_info.UbuntuDistroInfo().devel()
    except distro_info.DistroDataOutdated:
        devel = distro_info.UbuntuDistroInfo().stable()

    lts = distro_info.UbuntuDistroInfo().lts()
```

paride: Now I want again to implement it in pycludlib. It shouldn't be difficult really.

paride: Mmh no, I don't want to, as python3-distro-info is too Debian/Ubuntu specific, and I don't think we want the library to depend on it. It's not even in PyPI (`distroinfo` is a different thing).

powersj: yeah for pycloudlib I would be -1 on it, but for our test scripts it can definitely include it.

paride: Done (a bit differently).

PR #33, opened 2019-09-24
Revert "bootspeed.sh: do not call `systemd-analyze plot`"

paride:
This reverts commit 841f1b672828ff7a5d7de4865853159844499f98.

The collection of the `systemd-analyze plot` SVGs was disabled because
they often take more space than all the other measurement artifacts
combined, and we are not immediately using them. However it may be
worth collecting them to ease the analysis of the boot process.


raharper:
Thanks, this will be useful.

PR #36, opened 2020-06-15
Sleep more before checking the cloud-init copr tasks

paride:
Sometimes 3 minutes are not enough for the COPR infrastructure to show
the list of chroots cloud-init is being built in. Bump the sleep time
to 5 minutes.

### DIFF
--- a/cloud-init/copr_build.py
+++ b/cloud-init/copr_build.py
@@ -120,9 +120,13 @@ def main(srpm, copr_conf=DEFAULT_COPR_CONF, dev=None):
 
     # 2018-03-05: adding sleep to let builds ramp up
     # this after a {'state': ['Not a valid choice.']} exception
-    time.sleep(180)
+    reps = 6
+    for naptime in [5]*reps + [10]*reps + [30]*reps:
+        time.sleep(naptime)
+        tasks = get_build_tasks(build)
+        if tasks:
+            break
 
-    tasks = get_build_tasks(build)
     check_build_status(build, tasks)
     check_test_chroot(tasks)
     print()


cloud-init/copr_build.py
```diff
@@ -120,7 +120,7 @@ def main(srpm, copr_conf=DEFAULT_COPR_CONF, dev=None):
 
     # 2018-03-05: adding sleep to let builds ramp up
     # this after a {'state': ['Not a valid choice.']} exception
-    time.sleep(180)
+    time.sleep(300)
```
raharper, line 123: Is it worth switching to a backoff?
```
for timeout in [60, 120, 240, 360]:
      time.sleep(timeout)
```
or check more frequently:

```
for timeout in range(1, 6):
   time.sleep(60)
```

paride: In practice this speed up the job completion only when it fails, as the full build takes more than 5 minutes. But sounds reasonable. I think a condition we can wait wait for is `tasks.keys()` to be non-empty.

paride:
If `tasks` is still empty at the end of the `for` loop the failure mode is the same as before:

```
Traceback (most recent call last):
  File "./copr_build.py", line 146, in <module>
    main(ARGS.srpm, ARGS.copr_conf, ARGS.dev)
  File "./copr_build.py", line 130, in main
    check_test_chroot(tasks)
  File "./copr_build.py", line 30, in check_test_chroot
    raise Exception('chroot {} unexpectedly not in tasks ({})'.format(
Exception: chroot epel-8-x86_64 unexpectedly not in tasks ({})
```

cloud-init/copr_build.py
```diff
@@ -120,9 +120,12 @@ def main(srpm, copr_conf=DEFAULT_COPR_CONF, dev=None):
 
     # 2018-03-05: adding sleep to let builds ramp up
     # this after a {'state': ['Not a valid choice.']} exception
-    time.sleep(180)
+    for timeout in range(1, 10):
+        time.sleep(30)
```
raharper, line 124: Do we have any experience with how quickly tasks show up?  i hate waiting 30 seconds if we could have tasks quicker.

```
factor=6
for timeout in [5] * factor + [10] * factor + [30] * factor:
   time.sleep(timeout)
```
Still close to the 300 seconds but tries quicker sooner and the scales up it's timeout
```
>>> [5] * factor + [10] * factor + [30] * factor
[5, 5, 5, 5, 5, 5, 10, 10, 10, 10, 10, 10, 30, 30, 30, 30, 30, 30]
>>> sum([5] * factor + [10] * factor + [30] * factor)
270
```

cloud-init/copr_build.py
```diff
@@ -120,9 +120,12 @@ def main(srpm, copr_conf=DEFAULT_COPR_CONF, dev=None):
 
     # 2018-03-05: adding sleep to let builds ramp up
     # this after a {'state': ['Not a valid choice.']} exception
-    time.sleep(180)
+    for timeout in range(1, 10):
```
raharper, line 123: s/timeout/_/

You don't use timeout so pyflakes should complain about that.

paride: TIL that it doesn't: https://github.com/PyCQA/pyflakes/issues/285

cloud-init/copr_build.py
paride: Sometimes the tasks show up in just a few seconds, sometimes it takes minutes, I guess it depends on the load of the copr build infrastructure. I updated the PR with the solution you suggested.

PR #40, opened 2020-07-22
bootspeed: better support for arm64 ec2 instances

paride:


boot-speed/clouds/measure-cloud.py
```diff
@@ -96,12 +96,8 @@ def measure(self, datadir, instances=1, reboots=1):
             self.ssh_keypair_name = ec2.key_pair.name
         ec2.use_key(self.ssh_pubkey_path, self.ssh_privkey_path, self.ssh_keypair_name)
 
-        if self.inst_type.split(".")[0] == "a1":
-            if release == "xenial" or release == "bionic":
-                # Workaround for LP: #1832386
-                daily = ec2.released_image(release=release, arch="arm64")
-            else:
-                daily = ec2.daily_image(release=release, arch="arm64")
+        if self.inst_type.split(".")[0] in ["a1", "m6g"]:
```
blackboxsw, line 99: I think this is more than just m6g types:
```python
import pycloudlib
ec2 = pycloudlib.EC2()
instance_types =  ec2.client.describe_instance_types()['InstanceTypes']
arm_instance_types = [t['InstanceType'] for t in instance_types if 'arm64' in t['ProcessorInfo']['SupportedArchitectures']]
print(arm_instance_types)
['c6g.large', 'c6g.16xlarge', 'r6g.large', 'r6g.8xlarge', 'm6g.metal', 'm6g.xlarge', 'a1.xlarge', 'a1.medium', 'r6g.medium', 'c6g.12xlarge', 'c6g.xlarge', 'r6g.16xlarge', 'a1.large', 'r6g.4xlarge']
```
Should we also be adding c6g and r6g here too?

Also it might be nice to adopt pycloudlib in a future PR to dynamically construct this list in the future to make sure we don't miss out on new instance types.

paride: That's much better, I didn't know about `describe_instance_types()`.I'll fix it tomorrow.

blackboxsw: Good deal, if pulling in pycloudlib in the future, then we can use `arm_instance_types` local var directly too and avoid the splitting self.inst_type.split(".") for that check.

paride: Should look better now. The script already pulls in pycloudlib, so there was very little to adjust!

PR #42, opened 2020-08-07
uaclient: add helper script for ec2 cleanup of stale VMs and images

blackboxsw:
This script cleans up vms, vpcs, security groups and images.

It will be called by a jenkins job running on torkoal per the following PR:
https://github.com/canonical/server-jenkins-jobs/pull/134

### DIFF
--- a/ubuntu-advantage-client/ec2_cleanup.py
+++ b/ubuntu-advantage-client/ec2_cleanup.py
@@ -63,7 +63,9 @@ def clean_ec2(tag):
         client.delete_key_pair(KeyName=key['KeyName'])
 
     print('# searching for amis matching tag {}'.format(tag))
-    for image in resource.images.filter(Owners=['self']).all():
+    for image in resource.images.filter(
+        Owners=['self'], Filters=tag_filter
+    ).all():
         print('removing custom ami %s' % image.id)
         client.deregister_image(ImageId=image.id)
 
--- a/ubuntu-advantage-client/ec2_cleanup.py
+++ b/ubuntu-advantage-client/ec2_cleanup.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+"""Clean up all running EC2 instances, VPCs, and storage.
+
+Copyright 2018 Canonical Ltd.
+Joshua Powers <josh.powers@canonical.com>
+"""
+import boto3
+
+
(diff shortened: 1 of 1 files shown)


ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -0,0 +1,64 @@
...
+        for internet_gateway in vpc.internet_gateways.filter(Filters=tag_filter):
+            print('terminating internet gateway %s' % internet_gateway.id)
+            internet_gateway.detach_from_vpc(VpcId=vpc.id)
+            internet_gateway.delete()
+
+        print('terminating vpc %s' % vpc.id)
+        vpc.delete()
+
+    print('# searching for ssh keys')
+    key_name = {'Name': 'key-name', 'Values': [CI_TAG]}
```
lucasmoura, line 47: I think we should the `tag` parameter here instead

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -0,0 +1,64 @@
...
+    for key in client.describe_key_pairs(Filters=[key_name])['KeyPairs']:
+        print('deleting ssh key %s' % key['KeyName'])
+        client.delete_key_pair(KeyName=key['KeyName'])
+
+    print('# searching for amis')
+    for image in resource.images.filter(Owners=['self']).all():
```
lucasmoura, line 53: Why is the images and snapshots not being deleted according to the tag ? It seems that we will delete all of them regardless of the tag associated to them

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: ahh yes I just saw that too. Will fix now

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: very good point will fix that now

blackboxsw: passed tag_filter to this `images.filter` function now too
Can test with something like `./ubuntu-advantage-client/.tox/behave-aws/bin/python3 ubuntu-advantage-client/ec2_cleanup.py -t uaclient-ci-*`

PR #43, opened 2020-08-10
uaclient: add helper script for azure cleanup

lucasmoura:
This script cleans up all resource associated with a resource group associated with a given tag.
It will be called by either a travis job at the end of each test run or through a separate daily jenkins job.

### DIFF
--- a/ubuntu-advantage-client/azure_cleanup.py
+++ b/ubuntu-advantage-client/azure_cleanup.py
@@ -5,6 +5,7 @@
 # Lucas Moura <lucas.moura@canonical.com>
 import os
 import json
+import sys
 
 from pycloudlib.azure.util import get_client
 from azure.mgmt.resource import ResourceManagementClient
@@ -15,7 +16,7 @@ import argparse
 CI_DEFAULT_TAG = "uaclient"
 
 
-def parse_args():
+def get_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "-t", "--tag", dest="tag", action="store", default=CI_DEFAULT_TAG,
@@ -50,7 +51,7 @@ def parse_args():
             """
         )
 
-    return parser.parse_args()
+    return parser
 
 
 def clean_azure(tag, client_id, client_secret, tenant_id, subscription_id):
@@ -96,7 +97,19 @@ def load_azure_config(credentials_file):
 
 
 if __name__ == '__main__':
-    args = parse_args()
+    parser = get_parser()
+    args = parser.parse_args()
+    individ_args = all(
+        [
+            args.client_id, args.client_secret,
+            args.tenant_id, args.subscription_id
+        ]
+    )
+    if not any([args.credentials_file, individ_args]):
+        print("Either --credentials-file or tenant and client args required")
+        parser.print_help()
+        sys.exit(1)
+
     if args.credentials_file:
         if not os.path.exists(args.credentials_file):
             raise Exception("File {} could not be found".format(
--- a/ubuntu-advantage-client/azure_cleanup.py
+++ b/ubuntu-advantage-client/azure_cleanup.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+"""Clean up all running Azure resources and resource groups."""
+
+# Copyright 2020 Canonical Ltd.
+# Lucas Moura <lucas.moura@canonical.com>
+
+from pycloudlib.azure.util import get_client
+from azure.mgmt.compute import ComputeManagementClient
+from azure.mgmt.resource import ResourceManagementClient
+
+import argparse
+
+
+CI_DEFAULT_TAG = "uaclient"
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-t", "--tag", dest="tag", action="store", default=CI_DEFAULT_TAG,
+        help=("Tag used to filter cloud resources for deletion. "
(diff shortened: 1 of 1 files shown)


ubuntu-advantage-client/azure_cleanup.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-t", "--tag", dest="tag", action="store", default=CI_DEFAULT_TAG,
+        help=("Tag used to filter cloud resources for deletion. "
+              "Default: {}".format(CI_DEFAULT_TAG))
+    )
+    parser.add_argument(
```
mitechie, line 24: it'd be good at some point if we could leverage a credential file vs args so that they weren't leaked in run logs and such.

ubuntu-advantage-client/azure_cleanup.py
```diff
@@ -0,0 +1,81 @@
...
+        ResourceManagementClient, config_dict
+    )
+
+    print('# searching for resource groups matching tag {}'.format(tag))
+    for resource_group in resource_client.resource_groups.list():
+        tags = resource_group.tags
+
+        if tags:
+            for tag_value in tags.values():
+                if tag_value.startswith(tag):
```
mitechie, line 62: this is a prefix match vs a "Tag used to filter cloud resources for deletion." I think it's worth being very clear and consistent if it's a prefix vs an exact match as these types of mistakes tend to lead to unexpected behavior.

ubuntu-advantage-client/azure_cleanup.py
```diff
@@ -0,0 +1,81 @@
...
+                    resource_group_name = resource_group.name
+                    print('# deleting resource group: {}'.format(
```
mitechie, line 64: This is good to note it's deleting but I'd also prefer if it output the match that led to the decision to delete. Maybe something like

    deleted resource group: XX with tag YY

or the like

ubuntu-advantage-client/azure_cleanup.py
lucasmoura: Got it, I will add that option

lucasmoura:
Thanks for the review @mitechie. I have updated the script with your suggestions. I will also check how we will use the credentials in the jenkins job

ubuntu-advantage-client/azure_cleanup.py
```diff
@@ -0,0 +1,109 @@
...
+    parser.add_argument(
+        "--tenant-id", dest="tenant_id", 
+        help="Tenant id used to access azure api"
+    )
+    parser.add_argument(
+        "--subscription-id", dest="subscription_id", 
+        help="Subscription id used to access azure api"
+    )
+    parser.add_argument(
+        "--credentials-path", dest="credentials_path",
```
mitechie, line 43: s/path/file. A path ends in a / and is a directory but this is intended to be a file location.

ubuntu-advantage-client/azure_cleanup.py
```diff
@@ -0,0 +1,109 @@
...
+
+def load_azure_config(credentials_path):
+    with open(credentials_path, 'r') as f:
+        return json.load(f)
+
+
+if __name__ == '__main__':
+    args = parse_args()
+    if args.credentials_path:
+        if os.path.exists(args.credentials_path):
```
mitechie, line 94: I generally prefer if a script handles the error of a path that does not exist directly but I guess for this purpose it's low risk.

lucasmoura: This is just an attempt to fallback to the raw credentials if necessary. But thinking of the error we will generate, it will not be clear for an user that the script failed because the path does not exist. I will update the script to raise an Exception if the file does not exists

lucasmoura:
@mitechie Updated the code with an exception for the case of a credential file that does not exist

ubuntu-advantage-client/azure_cleanup.py
```diff
@@ -0,0 +1,112 @@
...
+                    result = resource_client.resource_groups.delete(
+                        resource_group_name=resource_group_name
+                    )
+                    result.wait()
+                    break
+
+
+def load_azure_config(credentials_file):
+    with open(credentials_file, 'r') as f:
+        return json.load(f)
```
blackboxsw, line 88: I think we might need a mapping here because the typical .azure/credentials.json on a system looks like the following:
```json
{
  "clientId": "da49e81a-<REDACTED>",
  "clientSecret": "<REDACTED>-",
  "subscriptionId": "12aad61c-6de4-<REDACTED>",
  "tenantId": "a0fb9276-6cc8-4e7a-<REDACTED>",
  "activeDirectoryEndpointUrl": "https://login.microsoftonline.com",
  "resourceManagerEndpointUrl": "https://management.azure.com/",
  "activeDirectoryGraphResourceId": "https://graph.windows.net/",
  "sqlManagementEndpointUrl": "https://management.core.windows.net:8443/",
  "galleryEndpointUrl": "https://gallery.azure.com/",
  "managementEndpointUrl": "https://management.core.windows.net/"
}

```

We might need to only sources clientId, clientSecret, tenantId and subscriptionId  and map them to client_id, client_secret, tenant_id and subscription_id.

This way I can run this script as well from my own system which has previously run 'az login` without having to expand all the stored credentials in a different format from what az cli uses.

lucasmoura:
@blackboxsw Updated

PR #44, opened 2020-08-12
ua-client: ec2_cleanup add --before-date param to remove stale items

blackboxsw:
Allow the ability to pass --before-date parameter to ec2_cleanup.py to allow CI jobs to delete
any resource older than MM/DD/YY.

The --before-date param can take to format MM/DD/YY

To test:
```
pip3 install boto3
AWS_ACCESS_KEY_ID=<yours> AWS_SECRET_ACCESS_KEY=<your_key> python3 ubuntu-advantage-client/ec2_cleanup.py  -t uaclient-ci-* --before-date 08/11/20
```

### DIFF
--- a/ubuntu-advantage-client/ec2_cleanup.py
+++ b/ubuntu-advantage-client/ec2_cleanup.py
@@ -12,6 +12,13 @@ import re
 
 CI_DEFAULT_TAG = "uaclient-*"
 
+# Our CI pycloudlib use reuses a single shared VPC across multiple
+# CI jobs because VPCs counts are limited to 5 per region.
+# cloud-init has one vpc and uaclient a 2nd shared VPC.
+# If we have any instances running in this VPC, don't try to
+# Remote security_groups, subnets, gateways
+SHARED_VPC_TAG = "uaclent-integration"
+
 
 def parse_args():
     parser = argparse.ArgumentParser()
@@ -47,18 +54,23 @@ def delete_resource_by_tag(resource, tag, older_than_prefix):
 
     Returns: True if resource should be deleted
     """
-    for tag in resource.tags:
-        if tag["Key"] != "Name":
-            continue
-        tag_value = tag["Value"]
-        if older_than_prefix:
-            if tag_value >= older_than_prefix:  # Resource is newer
-                return False
-        if '*' in tag:
-            if not re.match(tag, tag_value):
-                return False  # Value !match the cmdline provided -t <regex>
-        elif tag_value != tag:
-            return False  # Value not equal the cmdline provided -t <value>
+    if isinstance(resource, dict):
+        if "KeyName" in resource:
+            tag_value = resource["KeyName"]
+    else:
+        for tag in resource.tags:
+            if tag["Key"] != "Name":
+                continue
+            tag_value = tag["Value"]
+
+    if older_than_prefix:
+        if tag_value >= older_than_prefix:  # Resource is newer
+            return False
+    if '*' in tag:
+        if not re.match(tag, tag_value):
+            return False  # Value !match the cmdline provided -t <regex>
+    elif tag_value != tag:
+        return False  # Value not equal the cmdline provided -t <value>
     return True
 
 
@@ -72,7 +84,8 @@ def clean_ec2(tag_prefix, older_than=None):
         "uaclient-integration")
     )
     tag_filter = [{'Name': 'tag:Name', 'Values': [tag_prefix]}]
-    vpc_filter = [{'Name': 'tag:Name', 'Values': ["uaclient-integration"]}]
+    vpc_filter = [{'Name': 'tag:Name', 'Values': [SHARED_VPC_TAG]}]
+    import pdb; pdb.set_trace()
     for vpc in list(resource.vpcs.filter(Filters=vpc_filter)):
         print('cleaning up vpc %s' % vpc.id)
         wait_instances = []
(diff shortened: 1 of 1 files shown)


ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
...
 
-    print('# searching for vpcs matching tag {}'.format(tag))
-    tag_filter = [{'Name': 'tag:Name', 'Values': [tag]}]
-    for vpc in list(resource.vpcs.filter(Filters=tag_filter)):
+    tag_time_prefix = get_tag_prefix(older_than)
+    print('# searching for vpcs matching tag {}'.format(
+        "uaclient-integration")
+    )
+    tag_filter = [{'Name': 'tag:Name', 'Values': [tag_prefix]}]
+    vpc_filter = [{'Name': 'tag:Name', 'Values': ["uaclient-integration"]}]
```
lucasmoura, line 65: I think we should still use the `tag_prefix`  to search vpc resources, right ?

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
...
+
+def clean_ec2(tag_prefix, older_than=None):
     """Clean up all running EC2 instances, VPCs, and storage."""
     client = boto3.client('ec2')
     resource = boto3.resource('ec2')
 
-    print('# searching for vpcs matching tag {}'.format(tag))
-    tag_filter = [{'Name': 'tag:Name', 'Values': [tag]}]
-    for vpc in list(resource.vpcs.filter(Filters=tag_filter)):
+    tag_time_prefix = get_tag_prefix(older_than)
```
lucasmoura, line 60: Shouldn't the `tag_time_prefix` be constructed regarding the `tag` parameter ?

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
...
+                "left instance %s running; newer than %s" % (
+                    inst.id, older_than
+                )
+            )
+        for inst in wait_instances:
+            print("waiting on %s" % inst.id)
+            inst.wait_until_terminated()
 
         for security_group in vpc.security_groups.filter(Filters=tag_filter):
+            if not should_delete_resource(security_group, tag_time_prefix):
```
lucasmoura, line 89: Maybe I am missing something, but I could not find the definition of `should_delete_resource`

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
         help=("Tag used to filter cloud resources for deletion. "
               "Default: {}".format(CI_DEFAULT_TAG))
     )
+    parser.add_argument(
+        "-o", "--older-than", dest="older_than", action="store",
```
mitechie, line 23: "older-than" I think of a number of days or something. e.g. older than 30 days or older than 1yr. What about hinting you're looking for a date with "before-date" and then removing on that? 

I also think the hh:mm:ss is noise in the sense of an automated cleanup script. Can we just keep the date and leave the time to be midnight of that date?

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
...
+        help=("Tag used to filter cloud resources for deletion. "
+              "Format: MM/DD/YY [HH:MM:SS].)"
+    )
     return parser.parse_args()
 
 
-def clean_ec2(tag):
+def get_tag_prefix(older_than):
```
mitechie, line 30: I think this needs a docstring about the older_than arg expected data type.

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
...
+    try:
+        time = datetime.datetime.strptime(older_than, "%m/%d/%y %H:%M:%S")
+        return time.strftime("uaclient-ci-%m%d-%H%M%S")
+    except ValueError:
+        time = datetime.datetime.strptime(older_than, "%m/%d/%y")
+        return time.strftime("uaclient-ci-%m%d")
+
+
+def is_resource_stale(resource, tag_time_prefix):
+    """Return True if the given resource is older than the tag_time_prefix.
```
mitechie, line 43: so this is "tag_time_prefix" vs the method that's just "tag_prefix". Since it's more than the time should the argument just be "tag_prefix"?

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -19,62 +19,125 @@ def parse_args():
...
+    for tag in resource.tags:
+        if tag["Key"] != "Name":
+            continue
+        if tag["Value"] >= tag_time_prefix:
+            return False
+    return True
+
+
+def clean_ec2(tag_prefix, older_than=None):
     """Clean up all running EC2 instances, VPCs, and storage."""
```
mitechie, line 56: I think the older_than would need a docstring on expected value types.

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: Our CI in EC2 actually hardcodes a shared VPC tag uaclient-integration which because each region limits vpc count to 5 max. So, all ci jobs have to shared the same vpc. Also, all security_groups, route_tables, subnets under the vpc carry that same shared tag. I've adapted the logic to account for this, and not delete sg, routes, subnets or vpc if there are shared resources still in the vpc.

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: Yes it should . just pushed a change for this.

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: +1 done

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: thanks @mitechie.

sorry, this review was on an earlier push of mine. I went through a bit of bikeshedding in my head and ended up with the following:
```
def delete_resource_by_tag(resource, tag, time_prefix):                         
```
which returns a bool True if the resource should be deleted based on tag or tag_time_prefix values.

Please take a look and see if you think that satisfies the intent of your concern here

ubuntu-advantage-client/ec2_cleanup.py
blackboxsw: +1 added docstring and changed to before-date as you suggested above.

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -4,77 +4,198 @@
...
+    return time.strftime(tag + "%m%d")
+
+
+def delete_resource_by_tag(resource, tag, time_prefix):
+    """Return whether the resource is older than time_prefix or has tag.
+
+    :param resource: Either a dict or boto3 instance related to a boto3
+        resource. This can be an instance, security_group, subnet etc. SSH keys
+        are processed as dictionaries which contain a KeyName key.
+
```
lucasmoura, line 63: I think we can remove this line

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -4,77 +4,198 @@
...
     """Clean up all running EC2 instances, VPCs, and storage."""
     client = boto3.client('ec2')
     resource = boto3.resource('ec2')
 
-    print('# searching for vpcs matching tag {}'.format(tag))
-    tag_filter = [{'Name': 'tag:Name', 'Values': [tag]}]
-    for vpc in list(resource.vpcs.filter(Filters=tag_filter)):
+    time_prefix = get_time_prefix(tag_prefix, before_date)
+    print('# searching for vpcs matching tag {}'.format(
+        "uaclient-integration")
```
lucasmoura, line 101: We can use the `SHARED_VPC_TAG` here

ubuntu-advantage-client/ec2_cleanup.py
```diff
@@ -4,77 +4,198 @@
...
+        if "KeyName" in resource:
+            tag_value = resource["KeyName"]
+    else:
+        for tag in resource.tags:
+            if tag["Key"] != "Name":
+                continue
+            tag_value = tag["Value"]
+
+    if time_prefix:
+        if tag_value >= time_prefix:  # Resource is newer
```
lucasmoura, line 84: Just to be sure here, if we have a tag like `uaclient-*`, time_prefix would be something like `uaclient-*08122020", right ? If that is true, I think this check will fail because of the "*" in the string.

blackboxsw: Got catch.

I fixed it now.

get_time_prefix("uaclient-*", "08/12/20") will now return 
"uaclient-0812"
instead of 

'uaclient-*-0812'

PR #48, opened 2020-09-16
Update azure_cleanup script

lucasmoura:
Right now, we are trying to use the `azure_cleanup` script in uaclient. However, this scrips was coded with the intention of removing all of the `resource_groups` at the end of all tests. Currently, we intend to run this script after each test has finished. To do that, we must updated how we perform the tag matching in this script


paride (approved):
LGTM

PR #56, opened 2020-10-08
Emit dots when waiting for azure cleanup on travis

lucasmoura:
Emit dots in the stdout when running the `azure_cleanup` scriprt on travis. Right now, we are using the `emit_dots_on_travis` function developed for `cloud-init` here [#347](https://github.com/canonical/cloud-init/pull/347)

An alternative here would be to import this function directly from `cloud-init`, but I don't know if we should assume that `cloud-init`will always be available when running this script


OddBloke (approved):
This looks correct to me, thanks!

PR #59, opened 2020-10-20
LMA: Don't grep for 'runtime-config' on cortex_test.sh

sergiodj:
The 'runtime-config' excerpt shows up in the output of
'$cortex_url/services' when we invoke cortex using the
'--runtime-config.file' option.  The runtime config file is something
we can provide to the cortex binary that allows us to change the
service configuration on-the-fly.

At the moment, we do not start cortex using this option.  For this
reason, it doesn't make sense for us to try to grep for this option in
the output.  This commit removes the grep, which makes the test
finally pass 100%.


paride (approved):
LGTM, thanks!

PR #60, opened 2020-10-21
OCI unit tests: postgresql

panlinux:
Before I add more tests following this pattern, let's do a review and see if we like where this is going.

The tests can be run in isolation:
```
cd lma-integration
sh tests/postgresql_test.sh
```
With extra debugging if so desired:
```
cd lma-integration
DEBUG_TESTS=1 sh tests/postgresql_test.sh
```
Or as part of the existing wrapper:
```
cd lma-integration
./run-tests
```
What else to factor out, or add to `setUp` and `tearDown`, will become clearer as more tests are added.
I also wonder if we shouldn't add the docker-compuse setup to a `oneTimeSetUp()` layer, or just move these unit tests to another directory as to keep these runs separate. There may be a smart way to put all this together using test suites. And/or maybe it's time to switch to python.


paride:
Hi @panlinux, it seems that the amd64 build of the Docker container is missing. I see only arm64 at https://hub.docker.com/r/squeakywheel/postgres/tags, while for example at https://hub.docker.com/r/squeakywheel/telegraf/tags I see amd64, arm64, ppc64el. Do we need to retrigger a build or something?

panlinux:
> Hi @panlinux, it seems that the amd64 build of the Docker container is missing. I see only arm64 at https://hub.docker.com/r/squeakywheel/postgres/tags, while for example at https://hub.docker.com/r/squeakywheel/telegraf/tags I see amd64, arm64, ppc64el. Do we need to retrigger a build or something?

https://bugs.launchpad.net/bugs/1900902 is affecting which builds are uploaded. If one build fails, none get uploaded.
The build failures are currently due to these bugs:
- arm64: [LP: #1891892](https://bugs.launchpad.net/rutabaga/+bug/1891892). Latest grafana build took more than 3h, and failed
- ppc64el: [LP: #1822632](https://bugs.launchpad.net/ubuntu/+source/gosu/+bug/1822632) gosu fails to run on ppc64el on bionic, focal

panlinux:
Odd that amd64 is gone for postgresql, I was running this test script all afternoon yesterday, pulling from squeakywheel/postgres:edge

panlinux:
Looks like I need `set -e` or some other way of tracking this, as the test now runs forever:
```
$ sh tests/postgresql_test.sh 
test_set_admin_user
Unable to find image 'squeakywheel/postgres:edge' locally
edge: Pulling from squeakywheel/postgres
docker: no matching manifest for linux/amd64 in the manifest list entries.
See 'docker run --help'.
ASSERT:
ASSERT:ERROR, failed to start container  in 10 seconds
ASSERT:ERROR, failed to start container  in 10 seconds
ASSERT:ERROR, failed to start container  in 10 seconds
ASSERT:ERROR, failed to start container  in 10 seconds
ASSERT:ERROR, failed to start container  in 10 seconds
...
```

paride:
@panlinux I'm not a great fan of `set -e` for non-trivial scripts, as it can easily cause more headaches than it solves. These are my standard references on it:

- https://mywiki.wooledge.org/BashFAQ/105
- https://lists.gnu.org/archive/html/bug-bash/2017-03/msg00170.html
- https://hackmd.io/IoVN08RrQV6Hyb6ztY4kwg (blog post from smoser)

In the specific case I'd explicitly `|| exit 1` or call a `fail()` function where necessary.

panlinux:
yeah, it's just a bug, as I had predicted this situation with the timeout limit of 10s. I'm just not failing when it's hit.

panlinux:
Ok, what caught me by surprise was that the assertions, fails, and similar failure cases did not end the test, they just flagged it as failed, but the test continued. So if I fail to start a container, obviously continuing running the test with a non-existent container will fail.

panlinux:
Ok, I think I addressed the problem where `fail()` and all the `assert*()` calls were not actually interrupting the test. Let me now try to use psql from a container, instead of the host.

panlinux:
@paride I might not have that last bit ready by tomorrow, so if you want to run the tests before that, please install `postgresql-client-common` on the host where you are running these tests, as that has the missing `psql` binary you showed me earlier today.

panlinux:
And new test instructions are:
```
$ sh oci-unit-tests/postgresql_test.sh 
test_set_admin_user
test_default_database_name
test_persistent_volume_keeps_changes

Ran 3 tests.

OK
```
Or, with extra debug:
```
$ DEBUG_TESTS=1 sh oci-unit-tests/postgresql_test.sh 
test_set_admin_user
Creating container with POSTGRES_USER=user180830
Waiting for container to be ready .....
Testing connection as user180830, looking for "postgres" DB
Trying to create a new DB called test_db180830 as user user180830
Verifying DB test_db180830 was created
test_default_database_name
Creating container with POSTGRES_DB=database180830
Waiting for container to be ready ......
Checking if database database180830 was created
test_persistent_volume_keeps_changes
Creating persistent volume
Launching container
Waiting for container to be ready ... 
Creating test database test_db_180830
Creating test table test_data_180830 with data
Launching new container with same volume
Waiting for container to be ready .
Verifying database test_db_180830 and table test_data_180830 are there with our data

Ran 3 tests.

OK
```

panlinux:
I can use psql from the postgresql image, but I'll have to create a docker network and launch all containers in that network in order to benefit from automatic dns setup. It's definitely not hard, but would make this branch a bit bigger. Basically:
```
docker network create dbtest
docker run --network dbtest ...... -d
docker run --network dbtest --rm --name psql squeakywheel/postgres:edge psql postgresql://postgres:secret@998be7a490ac/postgres -l
```
That hash-like hostname is the short container id (not the long one) of the postgresql server. The test script needs to either truncate the long one to that short version, or when launching the postgresql server container, give it a `--name` and use that name as the hostname instead of the hash. Both mechanisms work. And then manage removing the dbtest network at the end via a `tearDown` for each test, or a suite `tearDown`.

That's the plan, if you want it done in this branch, it's doable. But it can also be done in a separate branch after this lands.

sergiodj:
I think we can land this as-is, and then work on improving based on what @panlinux said.  I'd like to use this work as base for writing the unittests for redis.

paride:
Merging after MM + post-standup discussion on the caveats of this approach.

PR #62, opened 2020-11-05
OCI unit tests: Use psql from container

sergiodj:
This commit is based on Andrea's suggestion from PR #60.  Instead of
relying on the host's psql binary, we can (and indeed should!) use the
container's psql.  This has a bunch of positive points:

- It makes it unnecessary to install test dependencies on the host.

- It exercises the psql from inside the container, which is software
  that we will ship to the users anyway.

- It uses non-localhost hostnames for the tests (albeit we'd still
  be using a local docker network bridge).

The only "drawback" (if you can say so) is that it's not really
feasible to use here-document when invoking psql anymore, but this is
really not very significant for our purposes.

I've made sure the tests are still passing, and I don't have psql
installed in the host machine.

### DIFF
--- a/oci-unit-tests/postgresql_test.sh
+++ b/oci-unit-tests/postgresql_test.sh
@@ -8,10 +8,20 @@
 #  setUp() - run before each test
 #  tearDown() - run after each test
 
+DOCKER_NETWORK_NAME=postgresql_test
+
+oneTimeSetUp() {
+    docker network create $DOCKER_NETWORK_NAME > /dev/null 2>&1
+}
+
 setUp() {
     password=$(dd if=/dev/urandom bs=1 count=16 2>/dev/null | md5sum | head -c 16)
     id=$$
     image="squeakywheel/postgres:edge"
+}
+
+oneTimeTearDown() {
+    docker network rm $DOCKER_NETWORK_NAME > /dev/null 2>&1
 }
 
 tearDown() {
@@ -37,7 +47,9 @@ test_set_admin_user() {
 # it is not specified, then the default user of postgres will be used.
     admin_user="user${id}"
     debug "Creating container with POSTGRES_USER=${admin_user}"
-    container=$(docker run --rm -d \
+    container=$(docker run \
+        --network $DOCKER_NETWORK_NAME \
+        --rm -d \
         -e POSTGRES_USER=${admin_user} \
         -e POSTGRES_PASSWORD=${password} \
         -p 5432:5432 \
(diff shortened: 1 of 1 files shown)


sergiodj:
It's actually still possible to use here-documents with docker.  One just has to use docker's `-i` option:

```
cat << EOF | docker run --rm -i test/foo
 line1
 line2
EOF
```

I'll adjust the commit message accordingly.

paride:
@sergiodj LGTM, waiting for the adjusted commit message before merging.

oci-unit-tests/postgresql_test.sh
```diff
@@ -8,12 +8,22 @@
...
 setUp() {
     password=$(dd if=/dev/urandom bs=1 count=16 2>/dev/null | md5sum | head -c 16)
     id=$$
     image="squeakywheel/postgres:edge"
 }
 
+oneTimeTearDown() {
+    docker network rm $DOCKER_NETWORK_NAME
+}
+
```
panlinux, line 26: Have you checked that these setup and teardown work as expected? For example, induce a failure in a test later on, and check that the network is removed.

oci-unit-tests/postgresql_test.sh
```diff
@@ -48,16 +60,28 @@ test_set_admin_user() {
     wait_postgres_container_ready "${container}" || return 1
     debug "Testing connection as ${admin_user}, looking for \"postgres\" DB"
     # default db is still "postgres"
-    out=$(psql postgresql://${admin_user}:${password}@127.0.0.1 -q -l -A -t -F % | grep "^postgres" | cut -d % -f 1)
+    out=$(docker run \
+              --network $DOCKER_NETWORK_NAME \
+              --rm \
+              ${image} \
```
lucaskanashiro, line 66: You call `docker run` like this multiple times, I'd create a function for it. IMHO It would increase the readability of the tests.

oci-unit-tests/postgresql_test.sh
sergiodj: Yes, they work as expected.  If any test fails, the network still gets removed.

oci-unit-tests/postgresql_test.sh
sergiodj: I totally agree, but I'd like to do that in a separate PR.  IMO, the entire testcase can be improved by using more functions.

sergiodj:
@paride OK, commit message adjusted.

PR #63, opened 2020-11-06
OCI unit tests: redis

sergiodj:
This commit creates the unit tests for the redis OCI image.  It is
based on the postgresql unit tests, but with some
modifications/improvements.

Currently, it tests:

- A simple connection to a redis-server container without a password;

- A simple connection to a redis-server connection with a password;

- Insertion/retrieval of values to/from a redis-server container;

- Persistent volumes used with redis-server containers.

This is a good start, and we should work towards writing more tests
later.

### DIFF
--- a/oci-unit-tests/redis_test.sh
+++ b/oci-unit-tests/redis_test.sh
@@ -1,0 +1,195 @@
+. $(dirname $0)/helper/test_helper.sh
+
+# cheat sheet:
+#  assertTrue $?
+#  assertEquals 1 2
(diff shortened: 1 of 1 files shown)


paride (approved):
Tested locally, works fine, LGTM. There are some shellcheck warnings on the script (and on the postgresql one too), mostly about quoting, nothing actually unsafe in our specific use case. I'm for getting this running and then setup a CI job that does runs shellcheck on `oci-unit-tests/*.sh`.

PR #66, opened 2020-11-09
oci: add nginx unit test

paride:
Basic scenario covering the default configuration.


oci-unit-tests/nginx_test.sh
```diff
@@ -0,0 +1,118 @@
...
+#  assertTrue $?
+#  assertEquals 1 2
+#  oneTimeSetUp()
+#  oneTimeTearDown()
+#  setUp() - run before each test
+#  tearDown() - run after each test
+
+# The name of the temporary docker network we will create for the
+# tests.
+readonly DOCKER_NETWORK=nginx_test
```
lucaskanashiro, line 15: AFAICS we do not need this docker network, we are not using more than one container in these test cases. We should remove this variable, the creation and destruction of the network, and the `--network` parameter from the `docker run` call.

paride: I added a reverse proxy test that makes use of the Docker network :)

lucaskanashiro (approved):

PR #67, opened 2020-11-09
OCI unit test: prometheus

lucaskanashiro:
The following scenarios are tested:

- Call prometheus via command line.
- Check if the default target is up on online as it supposed to be.
- Check if data is persisted between containers destruction when using a
  docker volume. In this scenario passing a custom config file as a
  parameter was also exercised.

### DIFF
--- a/oci-unit-tests/prometheus_test.sh
+++ b/oci-unit-tests/prometheus_test.sh
@@ -1,0 +1,264 @@
+# shellcheck shell=dash
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+
+# cheat sheet:
+#  assertTrue $?
+#  assertEquals 1 2
+#  oneTimeSetUp()
+#  oneTimeTearDown()
+#  setUp() - run before each test
+#  tearDown() - run after each test
+
+# The name of the temporary docker network we will create for the
+# tests.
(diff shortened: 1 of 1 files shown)


paride (changes requested):
Noted a few things:

1. There is a bashism:
```
prometheus_test.sh: 82: eval: [[: not found
```
This won't work when executing the tests with `shunit2 <testfile>.sh`, as shunit2(1) suggests (broken anyway for other reasons, but I plan to fix it). In other words shunit2 seems to assume the script is sh-compliant. The other test scripts are already sh-compliant, I think we should stick with it.

2. The test tries to allocate a common port, which is already allocated on my system (by `docker-proxy`): 
```
docker: Error response from daemon: driver failed programming external connectivity on endpoint prometheus_test_409748 (a62e94e1183e278a8cb3a3c5dd1b7f02bac4dfd14b41e1c132c1d2534ec34b5f): Bind for 0.0.0.0:9090 failed: port is already allocated.
```
so the test fails

3. I think we should try to write shellcheck-clean files. In this case it's mostly a matter of stricter quoting. To let shellcheck know that sh is dash just add the following at the very first line of the script (see `nginx_test.sh`):
```
# shellcheck shell=dash
```

lucaskanashiro:
I addressed all your comments @paride , and also added some extra tests.

paride (approved):
LGTM!

PR #74, opened 2020-11-12
oci unit: add grafana

paride:


paride:
Not super easy to test given that actually testing requires interacting with the web UI. Suggestions welcome.

oci-unit-tests/grafana_test.sh
```diff
@@ -0,0 +1,107 @@
...
+
+    docker exec "$container" pgrep grafana-server > /dev/null
+    assertTrue $?
+
+    # Without Javascript the web UI doesn't do much. 
+    curl -Ss http://127.0.0.1:$LOCAL_PORT/login | grep -Fq "<title>Grafana</title>"
+    assertTrue $?
+
+    curl -Ss "http://[::1]:$LOCAL_PORT/login" | grep -Fq "<title>Grafana</title>"
+    assertTrue $?
```
sergiodj, line 63: IPv6 connectivity is already being tested by the function below, right?

oci-unit-tests/grafana_test.sh
```diff
@@ -0,0 +1,107 @@
...
+
+test_default_config() {
+    debug "Creating all-defaults Grafana container"
+    container=$(docker_run_server -p "$LOCAL_PORT:3000")
+
+    assertNotNull "Failed to start the container" "${container}" || return 1
+    wait_grafana_container_ready "${container}" || return 1
+
+    docker exec "$container" pgrep grafana-server > /dev/null
+    assertTrue $?
```
sergiodj, line 56: Out of curiosity, should we be using the `assertSomething foo || return 1` idiom whenever we call these assert functions?  That's what I'm doing for all my tests, but I noticed that the others tests don't always do that.

oci-unit-tests/grafana_test.sh
```diff
@@ -0,0 +1,107 @@
...
+    container=$(docker_run_server --user "$uid" -p "$LOCAL_PORT:3000" -v "$grafana_scratch:/var/lib/grafana")
+
+    assertNotNull "Failed to start the container" "${container}" || return 1
+    wait_grafana_container_ready "${container}" || return 1
+
+    curl -Ss http://127.0.0.1:$LOCAL_PORT/login | grep -Fq "<title>Grafana</title>"
+    assertTrue $?
+
+    debug "Shutting down container"
+    tearDown
```
sergiodj, line 91: I think it's better to call `stop_container_sync "${container}"` here, because the `tearDown` function is supposed to be called only when the test finishes (and it may do more than shutdown the container).

sergiodj (changes requested):
Thanks for the PR, @paride!

Overall it looks good, and I agree there's not much we can do other than resort to `curl` and check whether things are "running".  We could maybe write some test to use `lynx` and try to do some JavaScript processing, but I think it's OK to leave it for later.

When I run the tests here, I get this output:

```
test_default_config
test_default_config_ipv6
test_persistent_storage
(23) Failed writing body

Ran 3 tests.

OK
```

Do you know where this `(23) Failed writing body` is coming from?  It seems like it's something non-deterministic, because when I run the tests a second time it doesn't show up.

I've made a few comments in the code, and I think we will be good to merge this once they're addressed.

oci-unit-tests/grafana_test.sh
paride: Right, it's duplicate. I'll remove it from here.

oci-unit-tests/grafana_test.sh
paride: +1

oci-unit-tests/grafana_test.sh
paride: I prefer not to have the `|| return 1` and see all the following assertions fail (or succeed). With `|| return 1` the function returns right away, and we get less information. Plus we may have some test-specific teardown code after the asserts, which gets skipped if the function returns early. Still it could make sense in some cases I think.

sergiodj: Yeah, that's a good point.  I think we will have to come up with a standard way to write these assertions.

paride:
@sergiodj I also added the `docker pull`.

(I noticed that I pushed to the `canonical/server-test-scripts` instead of pushing to my fork. I'm trying the `gh` GitHub cli tool, which I recommend, but apparently it wants to be too smart! I will deleted the branch after it's merged.)

paride:
On `(23) Failed writing body`: I have no idea, and couldn't reproduce it locally :(

sergiodj:
> @sergiodj I also added the `docker pull`.
> 
> (I noticed that I pushed to the `canonical/server-test-scripts` instead of pushing to my fork. I'm trying the `gh` GitHub cli tool, which I recommend, but apparently it wants to be too smart! I will deleted the branch after it's merged.)

No problem!  BTW, I use Emacs and magit+forge are absolutely fantastic :-).

sergiodj:
> On `(23) Failed writing body`: I have no idea, and couldn't reproduce it locally :(

Oh, well.  Maybe it was just a fluke here.  Let's keep an eye and see if this happens again in production.  I'm approving this PR, thanks!

PR #75, opened 2020-11-13
OCI unit tests: telegraf

sergiodj:
Check whether:

- telegraf can properly start, and;

- telegraf can be invoked with a custom config file, and can write to
  an HTTP endpoint.

More tests should be added later.

### DIFF
--- a/oci-unit-tests/telegraf_test.sh
+++ b/oci-unit-tests/telegraf_test.sh
@@ -1,0 +1,90 @@
+# shellcheck shell=dash
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+
+# cheat sheet:
+#  assertTrue $?
+#  assertEquals 1 2
+#  oneTimeSetUp()
+#  oneTimeTearDown()
+#  setUp() - run before each test
+#  tearDown() - run after each test
+
+readonly DOCKER_PREFIX=oci_telegraf_test
+readonly DOCKER_IMAGE="squeakywheel/telegraf:edge"
+readonly TELEGRAF_PORT=9273
+
+oneTimeSetUp() {
+    # Cleanup stale resources
+    tearDown
+}
+
+setUp() {
+    suffix=$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 8 | head -n 1)
+}
+
+tearDown() {
+    for container in $(docker ps -a --filter "name=$DOCKER_PREFIX" --format "{{.Names}}"); do
+        debug "Removing container ${container}"
+        stop_container_sync "${container}"
+    done
+}
+
+# Helper function to execute telegraf.
+docker_run_telegraf() {
+    docker run \
+	   --rm \
+	   -d \
+	   --name "${DOCKER_PREFIX}_${suffix}" \
+	   "$@" \
+	   $DOCKER_IMAGE
+}
+
+wait_telegraf_container_ready() {
+    local container="${1}"
+    local log="\[agent\] Config: Interval:.+, Quiet:.+, Hostname:.+"
+    wait_container_ready "${container}" "${log}"
+}
+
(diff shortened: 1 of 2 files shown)


paride (changes requested):
`test_telegraf_up` fails for me: `wait_telegraf_container_ready()` reports the container didn't start in time. The function waits for the
```
"\[agent\] Config: Interval:.+, Quiet:.+, Hostname:.+"
```
regexp to match against the output of `docker logs`, which in my case is:
```
2020-11-13T14:21:51Z I! Starting Telegraf 1.15.2+ds1-0ubuntu1~20.04.1
2020-11-13T14:21:51Z I! Loaded inputs: processes swap system cpu disk diskio kernel mem
2020-11-13T14:21:51Z I! Loaded aggregators: 
2020-11-13T14:21:51Z I! Loaded processors: 
2020-11-13T14:21:51Z I! Loaded outputs: influxdb
2020-11-13T14:21:51Z I! Tags enabled: host=2b5b679f5cad
2020-11-13T14:21:51Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"2b5b679f5cad", Flush Interval:10s
2020-11-13T14:21:51Z W! [outputs.influxdb] When writing to [http://localhost:8086]: database "telegraf" creation failed: Post http://localhost:8086/query: dial tcp 127.0.0.1:8086: connect: connection refused
```
I think there's a racey condition in `wait_container_ready()`: it keeps checking the last (`tail -1`) line of `docker logs` against the pattern, but if more than 1 lines are printed in in that second then all but the last line are lost.

It could make sense to check for the pattern in the full `docker logs` output:

```
diff --git a/oci-unit-tests/helper/test_helper.sh b/oci-unit-tests/helper/test_helper.sh
index 87b3bd8..5554a1f 100644
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -54,8 +54,7 @@ wait_container_ready() {
     local max=${timeout}
 
     debug -n "Waiting for container to be ready "
-    logs=$(docker logs "${id}" --tail 1 2>&1)
-    while ! echo "${logs}" | grep -qE "${msg}"; do
+    while ! docker logs "${id}" 2>&1 | grep -qE "${msg}"; do
         debug -n "."
         sleep 1
                timeout=$((timeout-1))
@@ -63,7 +62,6 @@ wait_container_ready() {
             fail "ERROR, failed to start container ${id} in ${max} seconds"
             return 1
         fi
-        logs=$(docker logs "${id}" --tail 1 2>&1)
     done
     debug
 }
```

WDYT?

sergiodj:

Yesterday, while writing the test, I was experiencing the same failure you're seeing.  It happens because telegraf's default configuration instructs it to try to write to an influxdb instance, which doesn't exist in our case.  I fixed this while doing the tests and published a new telegraf OCI image, so I think you just have to do a `docker pull` on the image.

Having said that, I also had a similar idea about tweaking `wait_container_ready` and make it `grep` the full log.  My plan was to have a fourth argument (optional) that the caller could provide in order to say "I want you to `grep` the full log".  I was thinking that it may make sense to keep the function more strict by default, since it's been working fine for us so far.  But I wouldn't oppose a patch like the one you posted above.

paride:
I deleted my locally cached copy of the telegraf image and indeed it now succeeds, so I'm +1 for this. However in general I think we should remove the locally cached images before running the tests, to make sure we check against the latest remove image, and to make sure it's available.

paride (approved):

sergiodj:
> I deleted my locally cached copy of the telegraf image and indeed it now succeeds, so I'm +1 for this. However in general I think we should remove the locally cached images before running the tests, to make sure we check against the latest remove image, and to make sure it's available.

Agreed.  We should use the `oneTimeSetUp()` function for that.  I'll see about preparing a PR.

sergiodj:
See #76.

PR #76, opened 2020-11-13
OCI unit tests: Make sure we're using the latest OCI image

sergiodj:
In a discussion with Paride, we noticed that the OCI images can become
stale and are not automatically updated when there is a new version
available.  For this reason, it's better to call "docker pull
$DOCKER_IMAGE" every time we're starting a unittest.


paride (approved):
LGTM

PR #78, opened 2020-11-16
OCI unit test: fix telegraf test

sergiodj:
The telegraf unit test is unfortunately broken due to the way
assertTrue is being called.  I discovered this because the telegraf
OCI image was not configured by default to export the prometheus data
via its localhost:9273/metrics endpoint.

I fixed the image, and now I'm fixing the tests.  I'm also
implementing the idea discussed with Paride, which is to make
"wait_container_ready" accept a fourth argument allowing the caller to
grep for $msg in the full log message, and not only in the last line
of the log.

Last, but not least, I added descriptive messages to the assertTrue
tests.

### DIFF
--- a/oci-unit-tests/telegraf_test.sh
+++ b/oci-unit-tests/telegraf_test.sh
@@ -53,15 +53,20 @@ wait_telegraf_container_ready() {
 # Verify that telegraf is up and running.
 test_telegraf_up() {
     debug "Creating telegraf container"
-    container=$(docker_run_telegraf)
+
+    # We have to start the container using the host's network,
+    # otherwise we won't be able to connect to it.  This is because
+    # the default configuration makes telegraf bind to
+    # "localhost:9273", and we don't want to specify any custom
+    # configuration file here.
+    container=$(docker_run_telegraf --network host)
     wait_telegraf_container_ready "${container}" || return 1
 
     debug "Verifying that we can access the web endpoint"
     local metrics
-    metrics=$(curl --silent "http://localhost:${TELEGRAF_PORT}/metrics")
-    assertTrue echo "${metrics}" | grep -q diskio || return 1
-    assertTrue echo "${metrics}" | grep -q inodes || return 1
-    assertTrue echo "${metrics}" | grep -q cpu || return 1
+    metrics="$(curl --silent "http://localhost:${TELEGRAF_PORT}/metrics")"
+    assertTrue "Check if the go_info metric is present" "echo \"${metrics}\" | grep -qF go_info"
+    assertTrue "Check if the cpu metric is present" "echo \"${metrics}\" | grep -qF cpu"
 }
 
 # Verify that using a custom configuration file works, and that
@@ -84,10 +89,10 @@ test_telegraf_custom_config_http_endpoint() {
     local data
     # We have to use "timeout" here because otherwise "nc" will keep
     # listening to the port.
-    data="$(timeout 5s nc -q 0 -w 20 -l -p 8080)"
-    assertTrue echo "${data}" | grep -q diskio || return 1
-    assertTrue echo "${data}" | grep -q inodes || return 1
-    assertTrue echo "${data}" | grep -q cpu || return 1
+    data="$(timeout 20s nc -q 0 -w 20 -l -s 127.0.0.1 -p 8080)"
+    assertTrue "Check if the diskio metric is present" "echo \"${data}\" | grep -qF diskio"
+    assertTrue "Check if the inodes metric is present" "echo \"${data}\" | grep -qF inodes"
+    assertTrue "Check if the cpu metric is present" "echo \"${data}\" | grep -qF cpu"
 }
 
 load_shunit2
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -50,20 +50,17 @@ wait_container_ready() {
     local id="${1}"
     local msg="${2}"
     local timeout="${3:-30}"
-    local logs
-    local max=${timeout}
+    local max=$timeout
 
     debug -n "Waiting for container to be ready "
-    logs=$(docker logs "${id}" --tail 1 2>&1)
-    while ! echo "${logs}" | grep -qE "${msg}"; do
-        debug -n "."
-        sleep 1
-		timeout=$((timeout-1))
-        if [ "${timeout}" -le 0 ]; then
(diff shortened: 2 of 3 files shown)


oci-unit-tests/telegraf_test.sh
```diff
@@ -47,21 +47,21 @@ docker_run_telegraf() {
...
     local log="\[agent\] Config: Interval:.+, Quiet:.+, Hostname:.+"
-    wait_container_ready "${container}" "${log}"
+    wait_container_ready "${container}" "${log}" 30 true
 }
 
 # Verify that telegraf is up and running.
 test_telegraf_up() {
     debug "Creating telegraf container"
-    container=$(docker_run_telegraf)
+    container=$(docker_run_telegraf --network host)
```
lucaskanashiro, line 56: This test seems to be depending on the host different from the other tests. I think we should document this better since this is an exception. Maybe add a comment on top of it explaining the why?

sergiodj: Yeah, I put a comment in the test below this one explaining why we're using the host network, but now that this test also depends on the host, I think it's fair to replicate the comment.  That's what I did.

oci-unit-tests/telegraf_test.sh
```diff
@@ -47,21 +47,27 @@ docker_run_telegraf() {
 wait_telegraf_container_ready() {
     local container="${1}"
     local log="\[agent\] Config: Interval:.+, Quiet:.+, Hostname:.+"
-    wait_container_ready "${container}" "${log}"
+    wait_container_ready "${container}" "${log}" 30 true
```
paride, line 50: There is a bit of a mismatch between this `true` and the fact that `wait_container_ready()` check if the argument is wither `0` or anything else, including `false`. In other words one could easily assume that the opposite of `true` is `false` here, it `false` still count as true. I normally prefer to store `true` or `false` in a variable and then evaluate it with `if $variable`, so the builtin `true`/`false` are evaluated.

However in this specific case I'd try not to add another argument to `wait_container_ready`, but to replace the whole `while` loop with something like:
```
if ! docker logs "${id}" --follow | timeout $timeout grep -qE "${msg}"; then
    fail "ERROR, failed to start container ${id} in ${max} seconds"
    return 1
fi
```
I think this brings the best of both worlds: returns success on the first occurrence of `$msg` in the logs, waiting for more logs for the given time.

paride (changes requested):
In general LGTM, one comment on how we want for the container to be ready.

oci-unit-tests/telegraf_test.sh
sergiodj: Thanks for the review.  That is a good idea, but unfortunately it doesn't work because `--follow` never exits.  It is possible to put use `timeout` to invoke it, but then the tests take a long time to complete.

I decided to implement a not-so-radical idea: just get rid of `--tail`, and make sure we're `grep`ing the full contents of `docker logs`.  This should suffice for what we want.  Let me know what you think.

sergiodj:
I've fixed a few more tests that were failing due to missing quotes.

oci-unit-tests/telegraf_test.sh
paride: Interesting! I thought `docker logs | timeout grep` would exit because once the right hand of the pipe exits (because `grep` returns of the timeout is hit) the pipe is broken, and a `SIGPIPE` is generated.. This works with plain old tail:
```
$ # exits after 2 seconds
$ tail -f /etc/passwd | timeout 2 grep -q notmatching

$ # shows the signal
$ strace tail -f /etc/passwd | timeout 2 grep -q notmatching
[...]
--- SIGPIPE {si_signo=SIGPIPE, si_code=SI_TKILL, si_pid=3032673, si_uid=1000} ---
+++ killed by SIGPIPE +++
```
But apparently `docker logs` ignores `SIGPIPE`, and keeps waiting.

Moving `timeout` to the right hand side of the pipe seems to work:
```
# Exits immediately:
$ timeout 30 docker logs --follow 891534c67c0f | grep -q .
```
This is because now it's `timeout` that catches the `SIGPIPE`, and exits. So I think that
```
if ! timeout $timeout docker logs "${id}" --follow | grep -qE "${msg}"; then
    fail "ERROR, failed to start container ${id} in ${max} seconds"
    return 1
fi
```
should work, exiting immediately after a match. Note that it works only because the pipeline's RC is the one of the rightmost command (`grep`). It wouldn't work with bash and `shopt -o pipefail`.

In the end, I'm also OK with your solution, it's simple and gets the job done. I did some digging mostly because I was puzzled by the fact that `docker logs` didn't exit.

paride (approved):
LGTM but I'll leave the merge to you.

oci-unit-tests/telegraf_test.sh
sergiodj: Hm, are you perhaps using `docker logs` in a container that is constantly printing information?  The `SIGPIPE` only happens when the left-hand side tries to write to the pipe; this will trigger the signal, which tells the program that the pipe has been closed.  It seems to me that `docker logs --follow` doesn't use line-buffered output, because moving `timeout` to the left-hand side of the pipe doesn't work, at least not here.  I had tried that before, and after seeing that it doesn't work, I decided to implement the solution I suggested.  Could you repeat the test with a telegraf container, just to make sure you can reproduce the failure there?

What happens here after moving `timeout` to the left-hand side of the pipe is that we have to wait the full `$timeout` before the command finishes.  It will finish successfully, but it will take a long time.

This happens because the telegraf container starts very quickly, and then doesn't print anything else.  I did a quick test here with a MySQL container, which takes almost 1 minute to start and will print a bunch of lines in the log.  In this particular case, the `SIGPIPE` was correctly generated & caught, and `docker logs --follow | grep` returned very quickly.

I also tried to forcing a line-buffered output using `stdbuf -oL -eL` and `grep --line-buffered`, to no avail.

paride: I could reproduce the behavior you describe, and I think you're right in thinking the culprit buffering. The result is racey-like. In bash we could use process substitution:
```
timeout 30 grep -q <(docker run ...)
```
at least I think: I've got surprised already. In plain `sh` we'd need a named pipe and handling of the `docker run` PID (kill it on match/timeout, otherwise it will stay attached to the pipe indefinitely, same of my first `docker logs | timeout grep` idea). Totally not worth it.

Thanks for this discussion, I've got a couple of nice TILs!

PR #79, opened 2020-11-17
OCI unit test: Fix cortex test

sergiodj:
Now that the cortex image was fixed and it is possible to start the
service without having to specify a configuration file, we can adjust
the unit test so that we can test the image as is.

I took the opportunity and deleted the (now unused) configuration file
for cortex, and also properly quoted the assertTrue tests we're
running.


paride (approved):
LGTM, tests fine.

PR #80, opened 2020-11-25
oci telegraf: drop unnecessary netcat parameters

paride:


sergiodj:
+1, thanks for investigating the failure.

PR #81, opened 2021-01-04
uaclient: add lxc cleanup script which removes stale lxc vms

blackboxsw:
Will be called by server-jenkins-jobs in a daily cron job


ubuntu-advantage-client/lxc_cleanup.py
```diff
@@ -0,0 +1,61 @@
...
+        for instance in json.loads(result.stdout):
+            if instance["name"].startswith(args.prefix):
+                created_at = datetime.datetime.strptime(
+                    instance["created_at"].split("T")[0], "%Y-%M-%d"
+                )
+                if created_at < before_date:
+                    print("Deleting: {}".format(instance["name"]))
+                    try:
+                        run(
+                            ["lxc", "stop", instance["name"]],
```
lucasmoura, line 56: Can't we just run the `lxc delete --force`  instead of stopping and deleting the instance ?

paride: Maybe that's because of the `ZFS dataset is busy` problems we hit in the past when using `delete --force`? I think `delete --force` can be considered safe now, I'm +1 for using it.

paride:
LGTM in general. Maybe we can call it `lxd_cleanup.py` instead of `lxc_`, to make it clear it's a LXD thing and not bare LXC.

flake8 complains about a few things that should be very easy to fix:

```
lxc_cleanup.py:6:1: F401 'contextlib.contextmanager' imported but unused
lxc_cleanup.py:8:1: F401 'multiprocessing' imported but unused
lxc_cleanup.py:9:1: F401 'os' imported but unused
lxc_cleanup.py:12:1: F401 'sys' imported but unused
lxc_cleanup.py:13:1: F401 'time' imported but unused
lxc_cleanup.py:17:1: E303 too many blank lines (3)
lxc_cleanup.py:59:21: E722 do not use bare 'except'
```

ubuntu-advantage-client/lxc_cleanup.py
blackboxsw: It was to avoid the ZFS dataset is busy issue, if that's looking ok now, I'll just use `delete --force`

paride (approved):

PR #82, opened 2021-01-05
Add GCP cleanup script

lucasmoura:
Add script to clean GCP instances


paride (approved):
LGTM but I didn't test it with GCE creds.

PR #83, opened 2021-01-05
Make lxd_cleanup script run on python 3.6

lucasmoura:
Bionic is the release used on some of our Jenkins nodes.
Because of that, we need to make our scripts run on
python 3.6. This PR fixes that issue for the lxd cleanup script


ubuntu-server-builder (approved):
+1  Thanks for catching this

blackboxsw (approved):
+1 thanks for catching this @lucasmoura  We'll need to support Bionic for jenkins nodes. I don't think we have any running Xenial anymore

PR #84, opened 2021-01-07
oci nginx: sleep 10s to ensure the container is ready

paride:


sergiodj:
Thanks for the PR, @paride.

I know this is not the ideal solution, but I think it's fine for now.  We should revisit this when we have the time.

PR #85, opened 2021-01-21
Add try except block when deleting ec2 resources

lucasmoura:
Currently, we do not check for failures when deleting ec2 resources. This means that we will abort the cleanup
script if we fail to delete any resource. This is not ideal, since we should be trying to cleanup as much as we can on each
run. We are updating the script to do that

### DIFF
--- a/ubuntu-advantage-client/ec2_cleanup.py
+++ b/ubuntu-advantage-client/ec2_cleanup.py
@@ -8,6 +8,7 @@ import argparse
 import boto3
 import datetime
 import re
+import traceback
 
 
 CI_DEFAULT_TAG = "uaclient-ci-*"
@@ -110,8 +111,14 @@ def clean_ec2(tag_prefix, before_date=None):
                 skipped_instances.append(instance)
                 continue
             print('terminating instance %s' % instance.id)
-            instance.terminate()
-            wait_instances.append(instance)
+            try:
+                instance.terminate()
+                wait_instances.append(instance)
+            except Exception:
+                print("Failure on terminating instance: {}".format(
+                    instance.id))
+                print("Exception: \n{}".format(traceback.print_exc()))
+
 
         for inst in skipped_instances:
             print(
@@ -137,14 +144,24 @@ def clean_ec2(tag_prefix, before_date=None):
                 skipped_resources = True
                 continue
             print('terminating security group %s' % security_group.id)
-            security_group.delete()
+            try:
+                security_group.delete()
+            except Exception:
+                print("Failure on terminating security group: {}".format(
+                    security_group.id))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         for subnet in vpc.subnets.filter(Filters=vpc_filter):
             if not delete_resource_by_tag(subnet, SHARED_VPC_TAG, time_prefix):
                 skipped_resources = True
                 continue
             print('terminating subnet %s' % subnet.id)
-            subnet.delete()
+            try:
+                subnet.delete()
+            except Exception:
+                print("Failure on terminating subnet: {}".format(
+                    subnet.id))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         for route_table in vpc.route_tables.filter(Filters=vpc_filter):
             if not delete_resource_by_tag(
@@ -153,7 +170,12 @@ def clean_ec2(tag_prefix, before_date=None):
                 skipped_resources = True
                 continue
             print('terminating route table %s' % route_table.id)
-            route_table.delete()
+            try:
+                route_table.delete()
+            except Exception:
+                print("Failure on terminating route table: {}".format(
+                    route_table.id))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         for internet_gateway in vpc.internet_gateways.filter(
             Filters=vpc_filter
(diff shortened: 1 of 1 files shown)


blackboxsw (changes requested):
Thanks lucas, a couple syntax errors on every print("Exception: ...) missing trailing paren "). 
Also what do you think about printing the whole traceback on failures here instead of just the string repr of the Exception? Is that of value:

Here's a diff that uses the traceback module (and fixes the trailing parenthesis):
```diff
diff --git a/ubuntu-advantage-client/ec2_cleanup.py b/ubuntu-advantage-client/ec2_cleanup.py
index 04a31d3..77b65a4 100755
--- a/ubuntu-advantage-client/ec2_cleanup.py
+++ b/ubuntu-advantage-client/ec2_cleanup.py
@@ -8,6 +8,7 @@ import argparse
 import boto3
 import datetime
 import re
+import traceback
 
 
 CI_DEFAULT_TAG = "uaclient-ci-*"
@@ -113,10 +114,10 @@ def clean_ec2(tag_prefix, before_date=None):
             try:
                 instance.terminate()
                 wait_instances.append(instance)
-            except Exception as e:
+            except Exception:
                 print("Failure on terminating instance: {}".format(
                     instance.id))
-                print("Exception: \n{}".format(str(e))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
 
         for inst in skipped_instances:
@@ -145,10 +146,10 @@ def clean_ec2(tag_prefix, before_date=None):
             print('terminating security group %s' % security_group.id)
             try:
                 security_group.delete()
-            except Exception as e:
+            except Exception:
                 print("Failure on terminating security group: {}".format(
                     security_group.id))
-                print("Exception: \n{}".format(str(e))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         for subnet in vpc.subnets.filter(Filters=vpc_filter):
             if not delete_resource_by_tag(subnet, SHARED_VPC_TAG, time_prefix):
@@ -157,10 +158,10 @@ def clean_ec2(tag_prefix, before_date=None):
             print('terminating subnet %s' % subnet.id)
             try:
                 subnet.delete()
-            except Exception as e:
+            except Exception:
                 print("Failure on terminating subnet: {}".format(
                     subnet.id))
-                print("Exception: \n{}".format(str(e))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         for route_table in vpc.route_tables.filter(Filters=vpc_filter):
             if not delete_resource_by_tag(
@@ -171,10 +172,10 @@ def clean_ec2(tag_prefix, before_date=None):
             print('terminating route table %s' % route_table.id)
             try:
                 route_table.delete()
-            except Exception as e:
+            except Exception:
                 print("Failure on terminating route table: {}".format(
                     route_table.id))
-                print("Exception: \n{}".format(str(e))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         for internet_gateway in vpc.internet_gateways.filter(
             Filters=vpc_filter
@@ -188,19 +189,19 @@ def clean_ec2(tag_prefix, before_date=None):
             try:
                 internet_gateway.detach_from_vpc(VpcId=vpc.id)
                 internet_gateway.delete()
-            except Exception as e:
+            except Exception:
                 print("Failure on terminating internet gateway: {}".format(
                     internet_gateway.id))
-                print("Exception: \n{}".format(str(e))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
         if not skipped_resources:
             print('terminating vpc %s' % vpc.id)
             try:
                 vpc.delete()
-            except Exception as e:
+            except Exception:
                 print("Failure on terminating vpc: {}".format(
                     vpc.id))
-                print("Exception: \n{}".format(str(e))
+                print("Exception: \n{}".format(traceback.print_exc()))
 
 
     print('# searching for ssh keys matching tag {}'.format(tag_prefix))
@@ -211,10 +212,10 @@ def clean_ec2(tag_prefix, before_date=None):
         print('deleting ssh key %s' % key['KeyName'])
         try:
             client.delete_key_pair(KeyName=key['KeyName'])
-        except Exception as e:
+        except Exception:
             print("Failure on terminating ssh key: {}".format(
                 key['KeyName']))
-            print("Exception: \n{}".format(str(e))
+            print("Exception: \n{}".format(traceback.print_exc()))
 
 
     print('# searching for amis matching tag {}'.format(tag_prefix))
@@ -226,10 +227,10 @@ def clean_ec2(tag_prefix, before_date=None):
         print('removing custom ami %s' % image.id)
         try:
             client.deregister_image(ImageId=image.id)
-        except Exception as e:
+        except Exception:
             print("Failure on removing custom ami: {}".format(
                 image.id))
-            print("Exception: \n{}".format(str(e))
+            print("Exception: \n{}".format(traceback.print_exc()))
 
     print('# searching for snapshots matching tag {}'.format(tag_prefix))
     for snapshot in resource.snapshots.filter(OwnerIds=['self']).all():
@@ -241,7 +242,7 @@ def clean_ec2(tag_prefix, before_date=None):
         except:
             print("Failure on removing custom snapshot: {}".format(
                 snapshot.id))
-            print("Exception: \n{}".format(str(e))
+            print("Exception: \n{}".format(traceback.print_exc()))
 
 
 if __name__ == '__main__':
```

blackboxsw (approved):
LGTM!

PR #86, opened 2021-01-28
oci test_helper: bump timeout to 60s and add debugging printout

paride:


paride:
I tested the `docker ps` thing by setting a 1s timeout in a local run, and it prints the container list to stdout. Can be useful to understand what happened when the timeout is hit.

sergiodj:
Thanks, @paride.  I wonder how many more times we'll need to bump the timeout :-/.  Anyway, this makes sense to me, so I'm merging it (and I liked the `docker ps` part!)

PR #88, opened 2021-02-18
Mysql test cleanup

bryceharrington:
The mysql_test.sh script didn't work due to some syntax errors in HERE-docs; these are repaired and the operations refactored to simplify them by introducing a docker_mysql_execute() routine that takes a SQL string.  I opted to pass override parameters like MYSQL_USER via local env variables, as I found them reasonably convenient and self-documenting; there's other ways to handle this so if you don't like it we can use a different approach.

(I notice the postgresql test could use a similar refactoring, but left that as is to keep this PR more focused on the mysql test.)

The test was also failing due to starting before the mysql database service was fully ready; thank you for diagnosing that to needing to watch for the port number.  This includes that fix.

Some of the messages used with assertEquals() were incorrect, and in fixing that I noticed some tests weren't providing the explanations, and the "cheat sheet" was documenting the two-argument form, whereas the three-argument form was more common.  This is all now more consistent.

### DIFF
--- a/oci-unit-tests/alertmanager_test.sh
+++ b/oci-unit-tests/alertmanager_test.sh
@@ -5,7 +5,7 @@
 
 # cheat sheet:
 #  assertTrue $?
-#  assertEquals 1 2
+#  assertEquals ["explanation"] 1 2
 #  oneTimeSetUp()
 #  oneTimeTearDown()
 #  setUp() - run before each test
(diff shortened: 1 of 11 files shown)


bryceharrington:
(Rebased on current master.)

sergiodj:
LGTM, thanks for improving the usability of these functions.  Initially I felt a bit uneasy with the rename to `docker_mysql_execute`, while the other tests are using the generic `docker_run_*` names, but I guess that's fine since mysql is a special case anyway.

PR #89, opened 2021-02-19
OCI: Use the "ubuntu/" namespace to test the images

sergiodj:
The test scripts are still using the "squeakywheel/" namespace, which
was just a temporary place to test the images.  This commit updates
the scripts to test the actual released images.


paride (approved):
LGTM, thanks!

PR #90, opened 2021-02-22
oci prometheus: fix test script

paride:


paride:
This should make the job succeed again; tested locally.

sergiodj:
Thanks for the fix, @paride.

PR #91, opened 2021-02-22
Rename postgresql to postgres

sergiodj:
This (postgres) is the official name of our OCI image; use it
throughout the test (where applicable, of course).

### DIFF
--- a/oci-unit-tests/postgres_test.sh
+++ b/oci-unit-tests/postgres_test.sh
@@ -8,8 +8,8 @@
 #  setUp() - run before each test
 #  tearDown() - run after each test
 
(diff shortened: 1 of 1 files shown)


paride (approved):
Thanks!

cpaelzer:
TBH - you don't ship the Postgresql project but the postgresql database.
So using the latter everywhere would make more sense IMHO, but I guess it is too late for that right?
=> https://www.postgresql.org/about/

sergiodj:
@cpaelzer if you're referring to the image name, then yeah, too late for that :-/.  If you referring to the modifications I've made to the test script, I can certainly review them and submit a new PR if needed, although I don't think it's worth the trouble :-).

PR #92, opened 2021-03-09
oci: parametrize and cleanup the tests

paride:
Changes:
- oci: fully parametrize the OCI image to test
- oci: add missing shellcheck directives
- oci: make all the scripts shellcheck-clean

### DIFF
--- a/oci-unit-tests/alertmanager_test.sh
+++ b/oci-unit-tests/alertmanager_test.sh
@@ -14,7 +14,11 @@
 # The name of the temporary docker network we will create for the
 # tests.
 readonly DOCKER_PREFIX=oci_alertmanager_test
-readonly DOCKER_IMAGE="${DOCKER_IMAGE:-ubuntu/prometheus-alertmanager:edge}"
+readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
+readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
+readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-prometheus-alertmanager}"
+readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
+readonly DOCKER_IMAGE="$DOCKER_REGISTRY/$DOCKER_NAMESPACE/$DOCKER_PACKAGE:$DOCKER_TAG"
 readonly ALERTMANAGER_PORT=60001
 
 oneTimeSetUp() {
--- a/oci-unit-tests/apache2_test.sh
+++ b/oci-unit-tests/apache2_test.sh
@@ -14,7 +14,11 @@
 # The name of the temporary docker network we will create for the
 # tests.
 readonly DOCKER_PREFIX=oci_apache2_test
-readonly DOCKER_IMAGE="${DOCKER_IMAGE:-ubuntu/apache2:edge}"
+readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
+readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
+readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-apache2}"
+readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
+readonly DOCKER_IMAGE="$DOCKER_REGISTRY/$DOCKER_NAMESPACE/$DOCKER_PACKAGE:$DOCKER_TAG"
 readonly LOCAL_PORT=59080
 
 oneTimeSetUp() {
--- a/oci-unit-tests/apache2_test.sh
+++ b/oci-unit-tests/apache2_test.sh
@@ -2,6 +2,7 @@
 
 # shellcheck disable=SC1090
 . "$(dirname "$0")/helper/test_helper.sh"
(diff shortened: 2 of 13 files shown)


oci-unit-tests/alertmanager_test.sh
```diff
@@ -14,7 +14,11 @@
 # The name of the temporary docker network we will create for the
 # tests.
 readonly DOCKER_PREFIX=oci_alertmanager_test
-readonly DOCKER_IMAGE="${DOCKER_IMAGE:-ubuntu/prometheus-alertmanager:edge}"
+readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
+readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
+readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-prometheus-alertmanager}"
+readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
+readonly DOCKER_IMAGE="$DOCKER_REGISTRY/$DOCKER_NAMESPACE/$DOCKER_PACKAGE:$DOCKER_TAG"
```
sergiodj, line 21: Let's stick to using `${VARIABLE}` instead of `$VARIABLE`.

sergiodj (changes requested):
Thanks for the PR, @paride.

I'm leaving one comment regarding the use of `{}` to enclose the variable names.

I also have a suggestion.  Instead of repeating the pattern of having multiple variables set at the beginning of each script, I think we can simplify this code by putting these variables elsewhere (`helpers/vars.sh` or some such) and sourcing the file instead.

We have to treat `DOCKER_PACKAGE` in a special way because it will vary depending on the file we're in, but everything else can stay the same.  So we'd have this:

```bash
readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-$(basename $0 | sed 's@\(.*\)_test\.sh@\1@')}"
readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
readonly DOCKER_IMAGE="${DOCKER_REGISTRY}/${DOCKER_NAMESPACE}/${DOCKER_PACKAGE}:${DOCKER_TAG}"
```

This only works with bash (doesn't work with pure sh), but I think that's OK.

oci-unit-tests/alertmanager_test.sh
paride: Not that I'm against it, but why? We use the `$VARIABLE` syntax very often across these scripts, and I''m not aware of any recommendation about always using `${VARIABLE}`.

paride:
> Thanks for the PR, @paride.
> 
> I'm leaving one comment regarding the use of `{}` to enclose the variable names.
> 
> I also have a suggestion. Instead of repeating the pattern of having multiple variables set at the beginning of each script, I think we can simplify this code by putting these variables elsewhere (`helpers/vars.sh` or some such) and sourcing the file instead.

+1, makes sense.

> We have to treat `DOCKER_PACKAGE` in a special way because it will vary depending on the file we're in, but everything else can stay the same. So we'd have this:
> 
> ```shell
> readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
> readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
> readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-$(basename $0 | sed 's@\(.*\)_test\.sh@\1@')}"
> readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
> readonly DOCKER_IMAGE="${DOCKER_REGISTRY}/${DOCKER_NAMESPACE}/${DOCKER_PACKAGE}:${DOCKER_TAG}"
> ```
> 
> This only works with bash (doesn't work with pure sh), but I think that's OK.

All these scripts are sh (dash) compatible, and think we shouldn't break this. I think I can come up with something similar which which in plain sh.

paride:
Actually I think your example works fine with plain sh; what's the bashism you were referring to?

oci-unit-tests/alertmanager_test.sh
sergiodj: You're right that we also use `$VARIABLE` across the scripts.  My comment was more of a personal preference when we use environment variables as part of path components; I think it's more readable this way, TBH.

sergiodj:
> Actually I think your example works fine with plain sh; what's the bashism you were referring to?

```
$ sh telegraf_test.sh 
telegraf_test.sh: 79: ./helper/test_helper.sh: Syntax error: "(" unexpected (expecting "}")
```

sergiodj:
> > Actually I think your example works fine with plain sh; what's the bashism you were referring to?
> 
> ```
> $ sh telegraf_test.sh 
> telegraf_test.sh: 79: ./helper/test_helper.sh: Syntax error: "(" unexpected (expecting "}")
> ```

Ah, sorry, I just noticed that I have uncommited changes from another thing I was working on.  Of course, the error above is not related to the suggestion I made.

paride:
@sergiodj PR updated following your review comments.

sergiodj:
This is great, thanks.

PR #94, opened 2021-03-16
git-ubuntu: refactor for core20 base

basak:
Now that we're building on core20 based on a deb, much of the complexity
here can disappear. The base image is switched to Focal. snapcraft is
now always a snap, and so is lxd. The updated tests rely less on the
environment so git no longer needs configuring first.


paride (approved):
LGTM, thanks!

PR #95, opened 2021-03-16
git-ubuntu: configure the lp: shorthand in the VM

paride:


paride:
@basak with this we should be all set with the core20 based CI.

basak:
+1

PR #96, opened 2021-03-24
OCI: Fix apache2 configure loading

sergiodj:
The apache2 unit test is using $PWD when composing the path for the
configure files to be mounted inside the container.  This doesn't work
when we're running these tests from an external directory.  We should
use the same pattern we're using to source the helper scripts here,
with the addition of calling "realpath", which is necessary because
docker requires that mount paths be fullpaths.


paride (approved):
LGTM, thanks!

PR #97, opened 2021-03-24
OCI: Allow user to override $DOCKER_IMAGE

sergiodj:
I noticed that it's possible to override almost all common variables,
except for the DOCKER_IMAGE one.  I think this is useful when doing a
"quick-and-dirty" local test when you want to just use your locally
built image.


paride (approved):
LGTM, handy addition, thanks!

PR #98, opened 2021-03-24
WIP: OCI: Implement test to check if images have manifest files

sergiodj:
Given the requirement that the images must contain manifest
files (under /usr/share/rocks) for the Security team, it's a good idea
to extend our unit tests and make sure we confirm the presence of said
files.

I'd have preferred something a little less repetitive than this
change, but unfortunately the way we start each service is different,
so we can't generalize it.

**NOTE**: This PR is marked as **WIP** because currently not all images have the manifests.  We are working on that (and should be done soon).  However, the code itself is ready for review.


paride:
@sergiodj in general this LGTM, I agree it's repetitive but I can't see an obvious way to generalize the test enough. I take the WIP as a "do not merge".

Could you please re-target the PR to the `main` branch, due to the `s/master/main/` rename?

sergiodj:
@paride thanks for the review.  The PR is ready to be merged, but I thought that it didn't make sense to do so right now because otherwise we'd have a bunch of failing tests.  However, given that the email notifications have been temporarily disabled, I think it should be fine to do the merge.  I will retarget the PR and then we can proceed.  Thanks.

paride (approved):
+1

PR #99, opened 2021-03-25
Fix gcp cleanup

lucasmoura:
The parsing for gcp cleanup script is incorrect. It is parsing the month part of the string as minutes. We are now fixing that

### DIFF
--- a/ubuntu-advantage-client/gcp_cleanup.py
+++ b/ubuntu-advantage-client/gcp_cleanup.py
@@ -86,7 +86,7 @@ if __name__ == '__main__':
 
     if args.before_date:
         before_date = datetime.datetime.strptime(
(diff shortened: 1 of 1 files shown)


ubuntu-advantage-client/gcp_cleanup.py
```diff
@@ -86,7 +86,7 @@ def clean_gcp(credentials_path, project_id, tag, before_date, region, zone):
 
     if args.before_date:
         before_date = datetime.datetime.strptime(
-                    args.before_date, "%M/%d/%Y"
+                    args.before_date, "%m/%d/%Y"
```
blackboxsw, line 89: D'oh docs. +1

blackboxsw (approved):
LGTM silly capitalization

PR #100, opened 2021-04-09
Add initial version of HA testing scripts

lucaskanashiro:
The scope of this initial version is the `fence_virsh` agent only. There are scripts to set up a 3-nodes cluster, run a common test case, and then destroy the cluster VMs.

There is still room for improvements, so any suggestion is welcome. To run the tests some assumptions are made and they are documented in the `README.md` file:

- The host machine needs to have an user called `ubuntu` (we can change the name easily if we want) which should be able to run `virsh` commands.
- The public ssh key of the `ubuntu` user present in each node of the cluster needs to be added to `authorized_keys` file of the `ubuntu` user in the host. The cluster nodes need to be able to ssh into this user in the host and run `virsh` commands from there.

The steps above are not implemented in any of the scripts, I've been doing it manually. We need to discuss what would be the best solution to run this in our Jenkins nodes.


ha/create-vm.sh
```diff
@@ -0,0 +1,111 @@
...
+
+launch_vm() {
+  virt-install \
+  	--connect qemu:///system \
+  	--virt-type kvm \
+  	--name ${VM_NAME} \
+  	--ram ${RAM} \
+  	--vcpus=${VCPU} \
+  	--os-type linux \
+  	--os-variant ubuntu20.04 \
```
paride, line 89: `ubuntu20.04` is not available yet on the Bionic Jenkins nodes, but I think it's safe to specify `ubuntu18.04` as the os variant even when installing newer releases

ha/delete-cluster.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+virsh shutdown node01
+virsh shutdown node02
+virsh shutdown node03
```
paride, line 5: Let's go with `virsh destroy` here, as we don't care about cleanly shutting down the nodes. Also if the shutdown fails then the `virsh undefine` will also fail, leaving the nodes around.

ha/delete-cluster.sh
```diff
@@ -0,0 +1,10 @@
...
+
+sleep 20
```
paride, line 7: Using `virsh destroy` no arbitrary sleep should be needed here.

paride:
On the ssh key generation (follow-up from MM):

My idea for the ssh key management/generation is roughly the following. Basic ideas:L
- We use a fixed ssh keypair for each run of the test job
- The cluster setup script checks if the keypair exists and generates it if needed.
This could hopefully be done with something like:
1. Generate the keypair if it doesn't exist:
```
if [ ! -f $HOME/.ssh/virsh_fence_test_id_rsa ]; then
    ssh-keygen -q -f $HOME/.ssh/virsh_fence_test_id_rsa -N "" -C "virsh fence agent test key"
    cat $HOME/.ssh/virsh_fence_test_id_rsa >> $HOME/.ssh/authorized_keys
```
2. Start the nodes
3. Copy (scp/sftp) the `virsh_fence_test_id_rsa` keypair to the nodes as `~/.ssh/id_rsa`, so we don't need to use `ssh -i`.
4. Now the nodes should be able to ssh back to the host.

lucaskanashiro:
@paride please check if the last commit implements what you proposed above regarding ssh key management.

ha/setup-cluster.sh
```diff
@@ -0,0 +1,194 @@
...
+  fi
+}
+
+copy_ssh_key_to_all_nodes() {
+  run_in_all_nodes "mkdir -p /home/ubuntu/.ssh"
+  copy_to_all_nodes "$HOME/.ssh/virsh_fence_test_id_rsa"
+  run_in_all_nodes "cp /home/ubuntu/virsh_fence_test_id_rsa /home/ubuntu/.ssh/id_rsa"
+}
+
+setup_authorized_keys_in_all_nodes() {
```
paride, line 129: Do we actually need this? The host can ssh into the cluster nodes using the ssh key installed using the `user-data`. The nodes can ssh into the host because  `copy_ssh_key_to_all_nodes()` copied the `test_id_rsa` key to the nodes, and the public part of that keypair is in the host's `authorized_keys`.

Do we also need to add that pubkey to the nodes' authorized_keys?

ha/setup-cluster.sh
```diff
@@ -0,0 +1,194 @@
...
+  if [ ! -f $HOME/.ssh/virsh_fence_test_id_rsa ]; then
+    ssh-keygen -q -f $HOME/.ssh/virsh_fence_test_id_rsa -N "" -C "virsh fence agent test key"
+    cat $HOME/.ssh/virsh_fence_test_id_rsa.pub >> $HOME/.ssh/authorized_keys
+  fi
+}
+
+copy_ssh_key_to_all_nodes() {
+  run_in_all_nodes "mkdir -p /home/ubuntu/.ssh"
+  copy_to_all_nodes "$HOME/.ssh/virsh_fence_test_id_rsa"
+  run_in_all_nodes "cp /home/ubuntu/virsh_fence_test_id_rsa /home/ubuntu/.ssh/id_rsa"
```
paride, line 126: (To be verified) We may need to `chmod 600` the privkey, otherwise ssh will complain. At least IIRC, I didn't try this yet.

ha/setup-cluster.sh
lucaskanashiro: Right, I was adding the pubkey to the nodes' `authorized_key` to make my life easier to test things manually, so I can ssh into any node from another one. I think you are right and we do not need to call this function to run the automated tests.

ha/setup-cluster.sh
lucaskanashiro: I tried to ssh from a cluster node into my host and I had not problem with the current setup. But maybe it is safer to run the `chmod 600`.

ha/setup-cluster.sh
```diff
@@ -0,0 +1,183 @@
...
+
+CREATE_VM_SCRIPT="$(pwd)/create-vm.sh"
+
+VM01="node01"
+VM02="node02"
+VM03="node03"
+
+SSH="ssh -o StrictHostKeyChecking=no"
+SCP="scp -o StrictHostKeyChecking=no"
+
```
paride, line 16: I'd add a `check_requirements()` function that checks that all the dependencies are installed. Can be as simple as (untested):
```
check_requirements() {
    hash cmd1 cmd2 cmd3 || exit 127
}
```

lucaskanashiro:
@paride I think I addressed  all your comments, it is up for review again. I tested this branch in `dugtrio` and it works flawlessly, you just need to run `./run-tests.sh`. Any suggestion is still welcome :)

ha/setup-cluster.sh
```diff
@@ -0,0 +1,188 @@
...
+WORK_DIR="${1:-$(pwd)}"
+CONFIG_DIR="${2:-"${WORK_DIR}/config"}"
+
+CREATE_VM_SCRIPT="$(pwd)/create-vm.sh"
+
+VM01="node01"
+VM02="node02"
+VM03="node03"
+
+SSH="ssh -o StrictHostKeyChecking=no"
```
paride, line 14: Let's make this `SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR"` and the same for `sftp` to avoid noise in the logs.

ha/setup-cluster.sh
```diff
@@ -0,0 +1,188 @@
...
+
+loop() {
+  CMD="${1}"
+  until ${CMD}; do
+    sleep 30
+  done
+}
+
+block_until_cloud_init_is_done() {
+  for vm in "${VM01}" "${VM02}" "${VM03}"; do
```
paride, line 148: I think there is an easier way here: `run_in_all_nodes 'cloud-init status --wait'`. This will wait for cloud-init to finish before returning, on each node.

ha/setup-cluster.sh
```diff
@@ -0,0 +1,188 @@
...
+
+write_config_files() {
+  write_hosts
+  write_corosync_conf
+}
+
+run_in_all_nodes() {
+  CMD="${1}"
+  for node_ip in "${IP_VM01}" "${IP_VM02}" "${IP_VM03}"; do
+    ${SSH} ubuntu@${node_ip} ${CMD}
```
paride, line 109: Should we made this stricter? As in:
```
${SSH} ubuntu@${node_ip} ${CMD} || exit 1
```
and have the job fail if the command fails for whatever reason

ha/setup-cluster.sh
```diff
@@ -0,0 +1,188 @@
...
+  run_in_all_nodes "chmod 600 /home/ubuntu/.ssh/id_rsa"
+}
+
+copy_config_files_to_all_nodes() {
+  copy_to_all_nodes ${CONFIG_DIR}/hosts
+  copy_to_all_nodes ${CONFIG_DIR}/corosync.conf
+}
+
+
+loop() {
```
paride, line 140: We may be able to drop this function if we do `block_until_cloud_init_is_done` differently (see below)`

paride:
Should we put all these virsh-related scripts in a separate directory under `ha`, e.g. `ha/virsh`? Then when we implement the cloud fence agents tests then can go under `ha/clouds` if we manage to find/write a framework that covers several clouds, or `ha/<cloud>` if we end up writing a different set of scripts/tests for each cloud we want to test. I think the virsh tests will always be somehow different for the tests for "true" clouds.

ha/setup-cluster.sh
```diff
@@ -0,0 +1,188 @@
+#!/bin/bash
+
+set -x
+
+WORK_DIR="${1:-$(pwd)}"
+CONFIG_DIR="${2:-"${WORK_DIR}/config"}"
+
+CREATE_VM_SCRIPT="$(pwd)/create-vm.sh"
+
+VM01="node01"
```
paride, line 10: Let's namespace these a bit better, e.g. `VM01="fence-test-virsh-node01"`. It will be easier to remember what those VMs are about when doing `virsh list` or similar on the nodes :) Also namespacing the VMs like this allows scripts to know the VMs they "own" and can do destructive operations on (e.g. for cleanup).

ha/delete-cluster.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+virsh destroy node01
+virsh destroy node02
+virsh destroy node03
+
+virsh undefine node01 --remove-all-storage
+virsh undefine node02 --remove-all-storage
+virsh undefine node03 --remove-all-storage
```
paride, line 9: We should make this cleanup script idempotent so we can call it liberally to make sure the host is clean. I'd go with something like:
```
#!/bin/bash

VM_PREFIX="node0"

virsh list --state-running --name | grep "^$VM_PREFIX" | xargs -L 1 --no-run-if-empty virsh destroy
virsh list --state-shutoff --name | grep "^$VM_PREFIX" | xargs -L 1 --no-run-if-empty virsh undefine --remove-all-storage

# Check for leftover VMs (cleanup failed!)
! virsh list --all --name | grep -q "^$VM_PREFIX" || exit 1
```
with the PREFIX to be updated.

ha/tests/fence_virsh_test.sh
```diff
@@ -0,0 +1,72 @@
...
+oneTimeSetUp() {
+  IP_VM01=$(virsh domifaddr ${VM01} | grep ipv4 | xargs | cut -d ' ' -f4 | cut -d '/' -f1)
+  IP_VM02=$(virsh domifaddr ${VM02} | grep ipv4 | xargs | cut -d ' ' -f4 | cut -d '/' -f1)
+  IP_VM03=$(virsh domifaddr ${VM03} | grep ipv4 | xargs | cut -d ' ' -f4 | cut -d '/' -f1)
+}
+
+run_command_in_node() {
+  NODE="${1}"
+  CMD="${2}"
+  SSH="ssh -o StrictHostKeyChecking=no"
```
paride, line 13: Let's have `SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR"` here too

paride (changes requested):
Hi @lucaskanashiro, one last comment: could you make these scripts shellcheck-clean? It's mostly a matter of adding extra quotes I think.

For the rest I think this looks good! I left some inline comments, they should all be easy to address. If any isn't or doesn't look right to you let's rediscuss it.

Then I think this can land, and once landed I'll add a Jenkins job to run the test daily, using the new directory structure `ha/virsh` if you agree on that comment.

I have ideas for further improvements, but let's have a working version in place and then iterate over it. Thanks for all the work you did here!

ha/setup-cluster.sh
lucaskanashiro: I tried this approach before but it did not work for some reason. I think the current approach is correct and it will always work, won't it?

ha/setup-cluster.sh
lucaskanashiro: This was a workaround because `cloud-init status --wait` did not work for me.

lucaskanashiro:
@paride all the comments were addressed. I hope it is good to land now. FWIW I tested these changes in `dugtrio` and everything worked as expected.

paride (approved):
Thanks @lucaskanashiro! Let's land this and make some jobs spin.

PR #101, opened 2021-04-14
WIP: OCI: Test if all supported architectures have been published

sergiodj:
**NOTE: This is currently a WIP**

I've decided to take a stab at implementing the test to verify whether
all supported architectures of an image have been properly published.

I don't know if this is the best way to implement this; sometimes I
think it is, other times I think it would be better to write this test
in its own file.

The reason I chose to write this as a helper function (which then
means that it will have to be called by every `*_test.sh` script) is
because we are going to have the test matrix which will cover all
registries x all namespaces x all tags anyway, so why don't take
advantage of that?  But again, this could very well be placed in its
own test file; a benefit is that we can then write generic tests for
all images there (for example, another test that I'm writing now is
the one that verifies whether the published hashes on Dockerhub match
those published on AWS).  Ah, and yet another benefit would be that we
wouldn't need the ugly hack I had to do on `common_vars.sh`.  See?  I
told you, I haven't really decided what's best here :-P.

By the way, I know that we've mentioned during standup that we could
just test images from one registry and be done with it (since we will
have the other test in place which will guarantee that the images are
the same on Dockerhub and AWS), but I think it's actually better to
test everything, because the list of supported architectures may
differ between registries for the same image.

Either way, I'd like an opinion about the test itself and the approach
taken from you guys.  This currently only implements the test for
telegraf.  In order to see it running, you have to invoke it like:

```
$ DOCKER_REGISTRY=public.ecr.aws sh telegraf_test.sh
```

Bear in mind that:

* You will have to `docker login` into AWS first.
* The test will fail, because not all architectures are available.

### DIFF
--- a/oci-unit-tests/helper/common_vars.sh
+++ b/oci-unit-tests/helper/common_vars.sh
@@ -9,3 +9,6 @@ readonly DOCKER_IMAGE="${DOCKER_IMAGE:-${DOCKER_REGISTRY}/${DOCKER_NAMESPACE}/${
 
 readonly DOCKER_PREFIX="${DOCKER_PREFIX:-oci_${DOCKER_PACKAGE}_test}"
 readonly DOCKER_NETWORK="${DOCKER_NETWORK:-${DOCKER_PREFIX}_net}"
+
+# List of all supported architectures for the images.
+readonly SUPPORTED_ARCHITECTURES="amd64 arm64 ppc64el s390x"
--- a/oci-unit-tests/standalone-checks_test.sh
+++ b/oci-unit-tests/standalone-checks_test.sh
@@ -1,0 +1,101 @@
+# shellcheck shell=dash
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+. "$(dirname "$0")/helper/common_vars.sh"
+
+# cheat sheet:
+#  assertTrue $?
+#  assertEquals ["explanation"] 1 2
+#  oneTimeSetUp()
+#  oneTimeTearDown()
+#  setUp() - run before each test
+#  tearDown() - run after each test
+
+# Standalone checks that will be performed for each image.
+
+# Check that all supported architectures are present in the manifest
+# list for every image.
+test_all_supported_architectures_are_available()
+{
+    # We only support querying AWS for now.
+    local registries="aws"
+    local namespaces="ubuntu lts"
+    local utilsdir
+    local listed_architectures
+    local images_list
+    # Which tags to test.
+    local tags_regex='\(latest\|beta\|edge\)'
+    local ret=0
+
+    utilsdir="$(mktemp -d)/utils"
+    trap 'rm -rf ${utilsdir}' 0 INT QUIT ABRT PIPE TERM
+
+    debug "Checking if all supported architectures are present in the manifest list of each image"
+
+    if ! git clone -q --depth 1 https://git.launchpad.net/~canonical-server/ubuntu-docker-images/+git/utils "${utilsdir}"; then
+	debug "failed to clone 'utils' repository"
+	exit 1
+    fi
+
+    # Iterate over the available registries...
+    for registry in ${registries}; do
+        # ... and over the available namespaces...
+        for namespace in ${namespaces}; do
+            # ... and over the list of all images for the
+            # registry/namespace combination...
+            images_list=$("${utilsdir}"/list-all-images.sh \
+                                       --registry "${registry}" \
+                                       --namespace "${namespace}")
+            if [ "${DOCKER_PACKAGE}" != "$(basename "$0" | sed 's@\(.*\)_test\.sh@\1@')" ]; then
+                # If the user has specified an image name to be tested, then
+                # we just test it.
+                if ! images_list=$(echo "${images_list}" | grep -Fw "${DOCKER_PACKAGE}"); then
+                    continue
+                fi
+            fi
+            for image in ${images_list}; do
+                # ... and over the list of all available tags for each
+                # image...
+                for tag in $("${utilsdir}"/list-tags-for-image.sh \
+                                          --registry "${registry}" \
+                                          --namespace "${namespace}" \
+                                          --image "${image}" \
+                                 | grep "${tags_regex}"); do
+                    debug "Checking manifest list for ${namespace}/${image}:${tag} on ${registry}"
+                    # Obtain the manifest list for the current tag,
+                    # and filter out everything but the published
+                    # architectures.
+                    listed_architectures=$("${utilsdir}"/list-manifest-for-image-and-tag.sh \
+					                --registry "${registry}" \
+					                --namespace "${namespace}" \
+					                --image "${image}" \
+					                --tag "${tag}" \
+			                       | jq -r '.manifests[].platform.architecture')
+                    if [ -z "${listed_architectures}" ]; then
+                        # Check if we've gotten a valid output from
+                        # the commands above.  Albeit rare, it is
+                        # possible that the manifest list comes out empty.
+                        echo "E: Could not obtain manifest list for ${namespace}/${image}:${tag} on ${registry}" > /dev/stderr
+                        ret=1
+                        continue
+                    fi
+                    for arch in ${SUPPORTED_ARCHITECTURES}; do
+	                if ! echo "${listed_architectures}" | grep -Fwq "${arch}"; then
+	                    echo "E: architecture '${arch}' not found in the manifest list of ${namespace}/${image}:${tag} on ${registry}" > /dev/stderr
+	                    ret=1
+	                else
+                            debug "architecture '${arch}' successfully found in the manifest list of ${namespace}/${image}:${tag} on ${registry}"
+                        fi
+                    done
+                done
+            done
+        done
+    done
+
+    rm -rf "${utilsdir}"
+
+    assertTrue "Not all supported architectures are available" "$ret"
+}
+
+load_shunit2


oci-unit-tests/helper/common_vars.sh
```diff
@@ -1,11 +1,14 @@
...
-readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
-readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
-readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-$(basename "$0" | sed 's@\(.*\)_test\.sh@\1@')}"
-readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
-readonly DOCKER_IMAGE="${DOCKER_IMAGE:-${DOCKER_REGISTRY}/${DOCKER_NAMESPACE}/${DOCKER_PACKAGE}:${DOCKER_TAG}}"
+[ -z "${DOCKER_REGISTRY}" ] && readonly DOCKER_REGISTRY="${DOCKER_REGISTRY:-docker.io}"
+[ -z "${DOCKER_NAMESPACE}" ] && readonly DOCKER_NAMESPACE="${DOCKER_NAMESPACE:-ubuntu}"
+[ -z "${DOCKER_PACKAGE}" ] && readonly DOCKER_PACKAGE="${DOCKER_PACKAGE:-$(basename "$0" | sed 's@\(.*\)_test\.sh@\1@')}"
+[ -z "${DOCKER_TAG}" ] && readonly DOCKER_TAG="${DOCKER_TAG:-edge}"
+[ -z "${DOCKER_IMAGE}" ] && readonly DOCKER_IMAGE="${DOCKER_IMAGE:-${DOCKER_REGISTRY}/${DOCKER_NAMESPACE}/${DOCKER_PACKAGE}:${DOCKER_TAG}}"
```
paride, line 8: I am not sure on why you had to add this "if $VAR empty => set VAR to itself if not-empty-or-null, or fallback to default value". Is this the ugly hack you mentioned in the PR description? I'm sure you have a good reason for it, but hopefully there's a nicer way...

paride:
> but I think it's actually better to test everything, because the list of supported architectures may differ between registries for the same image.

I am a bit confused by this. Starting with the assumption that both the registries support all the target architectures (currently amd64, arm64, s390x, ppc64el), are you saying that for some images we're going to have images for different subsets of the archs in the two registries? E.g. amd64+arm64 on dockerhub, and amd64+s390x on AWS?

oci-unit-tests/helper/test_helper.sh
```diff
@@ -97,3 +100,61 @@ check_manifest_exists()
...
+	    ;;
+	*"aws"*)
+	    registry="aws"
+	    ;;
+    esac
+
+    utilsdir="$(mktemp -d)/utils"
+    trap 'rm -rf ${utilsdir}' 0 INT QUIT ABRT PIPE TERM
+
+    debug -n "Checking if all supported architectures are present in the Docker manifest file"
```
athos-ribeiro, line 130: Is this by any chance a manifest list or an OCI image index? The reason I am asking is due to the "Accept" headers being passed to curl in the `https://git.launchpad.net/~canonical-server/ubuntu-docker-images/+git/utils` scripts, and to the fields being parsed with `jq` later in this commit.

If so, should the naming be replaced to reflect we are using manifest lists (or image indexes) instead of image manifests for (possibly) fetching the actual manifest for each architecture? Of course, if this is the case, this would probably go into another PR since other places may also refer to the manifest lists as manifests.

See https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list and https://docs.docker.com/registry/spec/manifest-v2-2/#image-manifest for further reference.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -97,3 +100,61 @@ check_manifest_exists()
     debug "not found"
     return 1
 }
+
+# Check that the images listed
```
athos-ribeiro, line 104: Is there any typos in this comment? not sure if it is clear to me...

oci-unit-tests/telegraf_test.sh
```diff
@@ -104,4 +104,11 @@ test_manifest_exists() {
     assertTrue "Manifest file(s) do(es) not exist in image" $?
 }
 
+test_all_supported_architectures_are_available() {
+    debug "Testing that the supported architectures are available in the manifest"
+
+    check_supported_architectures
+    assertTrue "Some supported architectures are not present in the registry" $?
```
athos-ribeiro, line 111: Any chance we could list the missing arches in this test output and/or list the manifest list? This would likely aid debugging in the future.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -97,3 +100,61 @@ check_manifest_exists()
...
+	return 1
+    fi
+
+    listed_architectures=$("${utilsdir}"/list-manifest-for-image-and-tag.sh \
+					--registry "${registry}" \
+					--namespace "${DOCKER_NAMESPACE}" \
+					--image "${DOCKER_PACKAGE}" \
+					--tag "${DOCKER_TAG}" \
+			       | jq -r '.manifests[].platform.architecture')
+
```
paride, line 149: General comment: there is a mix of spaces and tabs in this change (and in  part in the already committed scripts). It's not stated anywhere, but ~all the other scripts in this repo are using a 4-space indent, no tabs.

/bikeshed

oci-unit-tests/helper/common_vars.sh
sergiodj: Yeah, this is the ugly hack.  It's needed because `test_helper.sh` will need to include `common_vars.sh`.  However, `common_vars.sh` is already being included by the `*_test.sh` scripts, which means that these variable will have been initialized already.  Since they're `readonly`, we get an error.

I have to say that I've basically decided to drop this PR and implement this in a single script, so we likely won't need this hack at all.

sergiodj:
> > but I think it's actually better to test everything, because the list of supported architectures may differ between registries for the same image.
> 
> I am a bit confused by this. Starting with the assumption that both the registries support all the target architectures (currently amd64, arm64, s390x, ppc64el), are you saying that for some images we're going to have images for different subsets of the archs in the two registries? E.g. amd64+arm64 on dockerhub, and amd64+s390x on AWS?

I'm not saying that this will happen for certain, but given that AWS supports a subset of the architectures supported by Dockerhub, we may have to take this into account.  But to answer your question directly, I'm not aware of any discrepancies right now, so I think I may have "jumped the gun" here...  Let's stick to the original plan of testing only the AWS images.

oci-unit-tests/helper/test_helper.sh
sergiodj: Ah, fair enough.  I think I'm the only one using Emacs here :-).  I'll fix the indentation.

oci-unit-tests/helper/test_helper.sh
sergiodj: Heh, not a typo per se, but I clearly forgot to finish the sentence.  Thanks.

oci-unit-tests/telegraf_test.sh
sergiodj: Yes, good point.

oci-unit-tests/helper/test_helper.sh
sergiodj: Ah, I wasn't aware of the difference.  Based on the reference you gave, we're working with a manifest list.  I will update the naming to reflect that.  Thank you.

paride:
> > > but I think it's actually better to test everything, because the list of supported architectures may differ between registries for the same image.
> > 
> > 
> > I am a bit confused by this. Starting with the assumption that both the registries support all the target architectures (currently amd64, arm64, s390x, ppc64el), are you saying that for some images we're going to have images for different subsets of the archs in the two registries? E.g. amd64+arm64 on dockerhub, and amd64+s390x on AWS?
> 
> I'm not saying that this will happen for certain, but given that AWS supports a subset of the architectures supported by Dockerhub, we may have to take this into account. But to answer your question directly, I'm not aware of any discrepancies right now, so I think I may have "jumped the gun" here... Let's stick to the original plan of testing only the AWS images.

I can see how this could happen in principle, but should it happen I think it would be better to handle the different combinations in the Jenkins matrix job definition. Let's wait for the actual issue to show up (if ever), then it won't be too difficult to handle IMO.

sergiodj:
OK, I've revamped everything.

This has now been implemented in its own file, which IMO proved indeed to be the best choice here.  The code is cleaner, easier to read, and (perhaps most importantly), the ugly hack is no more.

The test is still not passing, but I encourage everyone to give it a try.  BTW, I've noticed that the set of published architectures for some images on AWS is indeed different than the one on Dockerhub.  I think this is not done on purpose, and I've already contacted the right people to double check this.

sergiodj:
> The test is still not passing, but I encourage everyone to give it a try. BTW, I've noticed that the set of published architectures for some images on AWS is indeed different than the one on Dockerhub. I think this is not done on purpose, and I've already contacted the right people to double check this.

FWIW, this has been resolved.

athos-ribeiro (commented):
Nice work, Sergio!

I added a few nitpicks inline, but that's all they are.

Moreover, I realized that, when one of the requirements on the sourced scripts fail, namely, having aws-cli installed, the tests output a nice error line informing us of the error, but the shunit2 tests still succeed.

Finally, I could still not successfully run the tests. From your (old) comments, one should `$ DOCKER_REGISTRY=public.ecr.aws sh telegraf_test.sh`, but AFAICT, that is not using the new tests  (sorry if I missed something in the discussion).

paride:
Hi @sergiodj, this looks good! A couple of questions:

1. Now that we have read-only credentials, do you think it would be OK to generate an authorization token *once* for the two registries using the read-only creds, and use just that on Jenkins, without setting up e.g. the AWS profiles for the two namespaces on the Jenkins nodes?

2. Now the test iterates over all the images. What do you think about making it check a single image (to be specified as an argument or env variable)? We could then have a matrix job in Jenkins that covers all the images. This could probably make it easier to spot a problem with a specific image.

sergiodj:

>     1. Now that we have read-only credentials, do you think it would be OK to generate an authorization token _once_ for the two registries using the read-only creds, and use just that on Jenkins, without setting up e.g. the AWS profiles for the two namespaces on the Jenkins nodes?

I'm not sure if that'd be possible.  The scripts we have (`list-*.sh`, from the `utils` repository) make extensive use of the `aws` command, which, as far as I remember, doesn't take an authorization token as input.   We will still need to configure the AWS profiles in order to use the CLI application.

>     2. Now the test iterates over all the images. What do you think about making it check a single image (to be specified as an argument or env variable)? We could then have a matrix job in Jenkins that covers all the images. This could probably make it easier to spot a problem with a specific image.

I thought about that, and I agree that's a good idea.  I will look into implementing it :-).

sergiodj:
> Nice work, Sergio!

Thanks for the review, @athos-ribeiro.

> Moreover, I realized that, when one of the requirements on the sourced scripts fail, namely, having aws-cli installed, the tests output a nice error line informing us of the error, but the shunit2 tests still succeed.

Ah, right.  I will work on detecting this earlier in the test.

> Finally, I could still not successfully run the tests. From your (old) comments, one should `$ DOCKER_REGISTRY=public.ecr.aws sh telegraf_test.sh`, but AFAICT, that is not using the new tests (sorry if I missed something in the discussion).

Now that the test lives in its own file, you would have to execute it like:

```
$ sh standalone-checks_test.sh
```

oci-unit-tests/standalone-checks_test.sh
```diff
@@ -0,0 +1,91 @@
...
+	exit 1
+    fi
+
+
+    # Iterate over the available registries...
+    for registry in ${registries}; do
+        # ... and over the available namespaces...
+        for namespace in ${namespaces}; do
+            # ... and over the list of all images for the
+            # register/namespace combination...
```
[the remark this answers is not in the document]
sergiodj: Ops.  Fixed.

oci-unit-tests/standalone-checks_test.sh
```diff
@@ -0,0 +1,91 @@
...
+    local ret=0
+
+    utilsdir="$(mktemp -d)/utils"
+    trap 'rm -rf ${utilsdir}' 0 INT QUIT ABRT PIPE TERM
+
+    debug "Checking if all supported architectures are present in the manifest list of each image"
+
+    if ! git clone -q --depth 1 https://git.launchpad.net/~canonical-server/ubuntu-docker-images/+git/utils "${utilsdir}"; then
+	debug "failed to clone 'utils' repository"
+	rm -rf "${utilsdir}"
```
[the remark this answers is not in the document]
sergiodj: Probably not.  I've removed it.

oci-unit-tests/standalone-checks_test.sh
```diff
@@ -0,0 +1,91 @@
...
+            for image in $("${utilsdir}"/list-all-images.sh \
+                                        --registry "${registry}" \
+                                        --namespace "${namespace}"); do
+                # ... and over the list of all available tags for each
+                # image...
+                for tag in $("${utilsdir}"/list-tags-for-image.sh \
+                                          --registry "${registry}" \
+                                          --namespace "${namespace}" \
+                                          --image "${image}" \
+                                 | grep '\(latest\|beta\|edge\)'); do
```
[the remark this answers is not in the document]
sergiodj: It could be, yeah.  Done.

oci-unit-tests/standalone-checks_test.sh
```diff
@@ -0,0 +1,91 @@
...
+                    debug "Checking manifest list for ${image}:${tag} on ${registry}"
```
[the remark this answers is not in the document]
sergiodj: Because we're not using the registry URL here, I opted to use the same format as I'm using in the other error message:

```
"E: Could not obtain manifest list for ${namespace}/${image}:${tag} on ${registry}"
```

oci-unit-tests/standalone-checks_test.sh
```diff
@@ -0,0 +1,91 @@
...
+					                --registry "${registry}" \
+					                --namespace "${namespace}" \
+					                --image "${image}" \
+					                --tag "${tag}" \
+			                       | jq -r '.manifests[].platform.architecture')
+                    if [ -z "${listed_architectures}" ]; then
+                        # Check if we've gotten a valid output from
+                        # the commands above.  Albeit rare, it is
+                        # possible that the manifest list comes out empty.
+                        echo "E: Could not obtain manifest list for ${namespace}/${image}:${tag} on ${registry}" > /dev/stderr
```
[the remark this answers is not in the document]
sergiodj: Ah, yes, I think it should.  Fixed.

sergiodj:
> > ```
> > 2. Now the test iterates over all the images. What do you think about making it check a single image (to be specified as an argument or env variable)? We could then have a matrix job in Jenkins that covers all the images. This could probably make it easier to spot a problem with a specific image.
> > ```
> 
> I thought about that, and I agree that's a good idea. I will look into implementing it :-).

I've re-read your suggestion, and I have a question for you.  Do you mean that you'd like this procedure to test a single image instead of iterating over the existing images?  Or would you like to still have the possibility of checking everything?

I've just pushed an implementation that supports specifying the image to be tested via the `DOCKER_IMAGE` variable, and defaults to test everything otherwise.  Let me know what you think.

athos-ribeiro (approved):

paride:
Hi @sergiodj, nice script, in general LGTM, however I have a few questions:

> I've just pushed an implementation that supports specifying the image to be tested via the `DOCKER_PACKAGE` variable, and defaults to test everything otherwise. Let me know what you think.

Is this on this branch? I can't see it implemented, and from this quick test:
```
$ DOCKER_PACKAGE=foobar AWS_CONFIG_FILE=$HOME/.aws/ecr_ro_config AWS_SHARED_CREDENTIALS_FILE=$HOME/.aws/ecr_ro_credentials bash standalone-checks_test.sh
test_all_supported_architectures_are_available
E: architecture 's390x' not found in the manifest list of ubuntu/grafana:7.4-21.04_edge on aws
[...]
```
it seems that the script still loops over all the images, not taking into account `DOCKER_PACKAGE=foobar`.

Another question: I many error lines like:
```
E: architecture 's390x' not found in the manifest list of ubuntu/grafana:7.4-21.04_edge on aws
E: architecture 's390x' not found in the manifest list of ubuntu/grafana:latest on aws
E: architecture 's390x' not found in the manifest list of ubuntu/grafana:7.4-21.04_beta on aws
E: architecture 's390x' not found in the manifest list of ubuntu/grafana:edge on aws
[...]
```
when running this script. Are the images actually missing? Is it expected at this stage?

Also I get error lines like:
```
An error occurred (AccessDeniedException) when calling the GetAuthorizationToken operation: User: arn:aws:iam::271607439846:user/ecr-ro is not authorized to perform: sts:GetServiceBearerToken on resource: *
E: Could not obtain manifest list for lts/memcached:1.5-20.04_edge on aws
```
is it expected at this stage, or should I triple-check my credentials?

(One nit: there's a mix of tabs and spaces in the indentation.)

sergiodj:
> Hi @sergiodj, nice script, in general LGTM, however I have a few questions:

> > I've just pushed an implementation that supports specifying the image to be tested via the `DOCKER_PACKAGE` variable, and defaults to test everything otherwise. Let me know what you think.
> 
> Is this on this branch? I can't see it implemented, and from this quick test:
> 
> ```
> $ DOCKER_PACKAGE=foobar AWS_CONFIG_FILE=$HOME/.aws/ecr_ro_config AWS_SHARED_CREDENTIALS_FILE=$HOME/.aws/ecr_ro_credentials bash standalone-checks_test.sh
> test_all_supported_architectures_are_available
> E: architecture 's390x' not found in the manifest list of ubuntu/grafana:7.4-21.04_edge on aws
> [...]
> ```
> 
> it seems that the script still loops over all the images, not taking into account `DOCKER_PACKAGE=foobar`.

Yeah, this is on this branch already.  It's strange that you're seeing it try to iterate over all images.  Here's what I see locally:

```
$ DOCKER_PACKAGE=foobar sh standalone-checks_test.sh 
test_all_supported_architectures_are_available

Ran 1 test.

OK
```

You can see that it didn't try any image.  (Arguably the script should error because the image name is invalid, but that's another problem).

> Another question: I many error lines like:
> 
> ```
> E: architecture 's390x' not found in the manifest list of ubuntu/grafana:7.4-21.04_edge on aws
> E: architecture 's390x' not found in the manifest list of ubuntu/grafana:latest on aws
> E: architecture 's390x' not found in the manifest list of ubuntu/grafana:7.4-21.04_beta on aws
> E: architecture 's390x' not found in the manifest list of ubuntu/grafana:edge on aws
> [...]
> ```
> 
> when running this script. Are the images actually missing? Is it expected at this stage?

Yes, this is expected.  The kernel team has a fix for this, but it will take a few weeks until we see it being deployed in Launchpad.  Until then, we will not have s390x images for 21.04.

> Also I get error lines like:
> 
> ```
> An error occurred (AccessDeniedException) when calling the GetAuthorizationToken operation: User: arn:aws:iam::271607439846:user/ecr-ro is not authorized to perform: sts:GetServiceBearerToken on resource: *
> E: Could not obtain manifest list for lts/memcached:1.5-20.04_edge on aws
> ```
> 
> is it expected at this stage, or should I triple-check my credentials?

This is not expected.  I will talk to Eric about this problem.

> (One nit: there's a mix of tabs and spaces in the indentation.)

Sorry about that, I have reconfigured my Emacs to always use spaces, but it's outsmarting me and reverting back to TABs-and-spaces when the file is already following this indentation.  I will reindent this whole file.

paride:
> > it seems that the script still loops over all the images, not taking into account `DOCKER_PACKAGE=foobar`.
> 
> Yeah, this is on this branch already. It's strange that you're seeing it try to iterate over all images. Here's what I see locally:

Now it works, but I now reset my local branch to yours so I can't tell what was wrong. Most likely my fault.

> The kernel team has a fix for this, but it will take a few weeks until we see it being deployed in Launchpad. Until then, we will not have s390x images for 21.04.

Good that the fix is landing! I see there are other non-s390x missing manifest errors, e.g.:
```
E: architecture 'ppc64el' not found in the manifest list of ubuntu/mysql:edge on aws
E: architecture 'arm64' not found in the manifest list of ubuntu/mysql:8.0-21.04_beta on aws
E: architecture 'ppc64el' not found in the manifest list of ubuntu/mysql:8.0-21.04_beta on aws
E: architecture 'arm64' not found in the manifest list of ubuntu/mysql:latest on aws
E: architecture 'ppc64el' not found in the manifest list of ubuntu/mysql:latest on aws
E: architecture 'amd64' not found in the manifest list of ubuntu/postgres:12-20.04_edge on aws
E: architecture 'amd64' not found in the manifest list of ubuntu/redis:6.0-21.04_beta on aws
E: architecture 'ppc64el' not found in the manifest list of ubuntu/redis:6.0-21.04_beta on aws
E: architecture 'amd64' not found in the manifest list of ubuntu/redis:6.0-21.04_edge on aws
E: architecture 'ppc64el' not found in the manifest list of ubuntu/redis:6.0-21.04_edge on aws
E: architecture 's390x' not found in the manifest list of ubuntu/redis:6.0-21.04_edge on aws
```
Are these also expected?

Anyway this PR lgtm now. Do you still consider it a WIP (per PR title), or do you think it's ready to land?

sergiodj:
Given that @paride and @athos-ribeiro reviewed and ACK'd the changes, I'm merging this PR.  Thanks!

PR #104, opened 2021-04-30
HA: add tests for IPaddr2 agent

lucaskanashiro:


paride:
The branch needs rebasing, but I tested it locally and `All tests successfully passed!`.

lucaskanashiro:
I did the rebase now it's updated. I needed to add the PPA again temporarily to be able to run tests against impish (`resource-agents` was not uploaded to the archive yet, it will soon).

paride:
This basically LGTM, but per MM discussion:
 - We'll keep this in the DRAFT state until `resource-agents-supported` reaches the archive, then we'll be able to test this against the real thing.
 - It would be nice to generalize some of the cluster management functions and have a `vm_utils.sh` library to be sourced in the various test scripts. Anyway I'm fine with merging this PR with some duplication and get jobs running, then we can improve with something to check against.

lucaskanashiro:
I pushed some improvements. Extracted some common functions and added it to `vm_utils.sh`.

The `resource-agents` package is already in `impish-proposed` but waiting in the NEW queue.

paride:
On top of the MM discussion: let's try to stick with a pattern on the test filenames. Now we have `tests/fence_virsh_test.sh` and `ipaddr2_test.sh`. We could name the latter `resource_ipaddr2_tests.sh` so the pattern is `AGENTTYPE_AGENTNAME_test.sh`.

paride (approved):
LGTM, thanks for addressing all the comments!

PR #107, opened 2021-05-06
ha/virsh/tests: add tests for the systemd resource agent

lucaskanashiro:
I still want to do some refactoring because there is some duplicated code in `resource_systemd_test.sh` and `resource_ipaddr2_test.sh`. But the logic of the tests is done, any feedback is welcome.


lucaskanashiro:
Now it is looking better and ready for review.

paride (approved):
LGTM, shellcheck clean, and I like the `test_helper.sh` split.

I tested the test and it tests fine :)

PR #108, opened 2021-06-04
OCI: Use "ppc64le" as the arch name

sergiodj:
Ref.: https://bugs.launchpad.net/bugs/1930094

The official architecture name that Docker uses is "ppc64le", not
"ppc64el".


paride (approved):
+1, thanks!

PR #110, opened 2021-06-07
OCI: Remove previous image before starting test

sergiodj:
When running a test whose image cannot be downloaded from the
registry, docker will use the cached image (from a previous run).  In
order to prevent this misleading scenario from happening, this commit
forces the removal of an existing image before we start testing.

Note that we just remove the image if `"${DOCKER_IMAGE}"` (which
contains the full image spec) starts with `"${DOCKER_REGISTRY}"`.  This
is necessary in order to continue supporting testing locally built
images.

### DIFF
--- a/oci-unit-tests/apache2_test.sh
+++ b/oci-unit-tests/apache2_test.sh
@@ -15,6 +15,9 @@
 readonly LOCAL_PORT=59080
 
 oneTimeSetUp() {
+    # Remove image before test.
+    remove_current_image
+
     # Make sure we're using the latest OCI image.
     docker pull --quiet "${DOCKER_IMAGE}" > /dev/null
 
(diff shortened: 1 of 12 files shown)


paride (commented):
Thanks @sergiodj, just one comment: as the image deletion operation is basically common code to all the images, should it go into a function defined in `helper/test_helper.sh`?

sergiodj:
> Thanks @sergiodj, just one comment: as the image deletion operation is basically common code to all the images, should it go into a function defined in `helper/test_helper.sh`?

Agreed, a function inside `helper/test_helper.sh` (which is called inside each `oneTimeSetUp`) looks better indeed.  I've updated the PR to reflect these changes.  Let me know what you think.

paride (approved):
LGTM!

PR #115, opened 2021-06-14
OCI nginx: use a dedicated network for test_reverse_proxy

paride:
This partially reverts f46faddec1b62ea5e74383d9c7ebc696b528fe8e,
adding back the creation/removal of DOCKER_NETWORK but using it
only for test_reverse_proxy, where it's actually needed.

For the rationale on why we're not using DOCKER_NETWORK globally see
the f46faddec1b62ea5e74383d9c7ebc696b528fe8e commit message.

### DIFF
--- a/oci-unit-tests/nginx_test.sh
+++ b/oci-unit-tests/nginx_test.sh
@@ -21,6 +21,14 @@ oneTimeSetUp() {
 
     # Cleanup stale resources
     tearDown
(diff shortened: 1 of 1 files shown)


sergiodj:
Hm, I wonder why the test has passed locally without the dedicated network.  Anyway, LGTM.  Thanks!

PR #116, opened 2021-06-14
OCI: Adjust message when number of images differ between registries

sergiodj:
Currently we print a message tagged as "error" when there is a
mismatch between the number of images available in one registry and
another.  We also print a misguided message saying that the test will
be aborted, when in fact it is not.

This message can be just a warning, since it doesn't actually impact
the actual test.

### DIFF
--- a/oci-unit-tests/standalone-checks_test.sh
+++ b/oci-unit-tests/standalone-checks_test.sh
@@ -181,17 +181,13 @@ test_images_hashes_are_equal()
                 cnt_images_list_tmp=$(echo "${images_list_tmp}" | wc -w)
 
                 if [ "${cnt_images_list}" -lt "${cnt_images_list_tmp}" ]; then
-                    echo "E: Registry '${registry}' contains more images than '${orig_registry}'.  Here are the extra images:" > /dev/stderr
+                    echo "W: Registry '${registry}' contains more images than '${orig_registry}'.  Here are the extra images:" > /dev/stderr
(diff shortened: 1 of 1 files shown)


oci-unit-tests/standalone-checks_test.sh
```diff
@@ -181,17 +181,13 @@ test_images_hashes_are_equal()
                 cnt_images_list_tmp=$(echo "${images_list_tmp}" | wc -w)
 
                 if [ "${cnt_images_list}" -lt "${cnt_images_list_tmp}" ]; then
-                    echo "E: Registry '${registry}' contains more images than '${orig_registry}'.  Here are the extra images:" > /dev/stderr
+                    echo "W: Registry '${registry}' contains more images than '${orig_registry}'.  Here are the extra images:" > /dev/stderr
```
paride, line 184: Out of curiosity: do you have a reason to prefer `> /dev/stderr` over the more common `>&2`? My guess it: `> /dev/stderr` is more explicit and we don't care about portability to old/obscure systems, but I wonder if there's more :)

paride (approved):
+1, thanks!

oci-unit-tests/standalone-checks_test.sh
sergiodj: No particular reason :-).  I'm just used to write code like this, so I just don't think about other ways nowadays ;-).

PR #117, opened 2021-06-25
HA: iscsi setup

lucaskanashiro:
The proposed changes add a `--iscsi` option to `setup-cluster.sh`, it will create a separate VM to be used as an iscsi target (in the future it could be used for other external services) and configure all the cluster nodes as iscsi initiators. If this new option is not passed to the script the old behavior will be the same.

If you want to test you can run `setup-cluster.sh --iscsi`, ssh into any cluster node and check if the iscsi block device is there:

```
ubuntu@fence-test-virsh-impish-node01:~$ sudo lsblk --scsi
NAME HCTL       TYPE VENDOR   MODEL         REV SERIAL                               TRAN
sda  6:0:0:0    disk LIO-ORG  iscsi-disk01 4.0  2f87b5b9-7f8a-49d3-9f85-d917a48c5c66 iscsi
ubuntu@fence-test-virsh-impish-node01:~$ cat /proc/partitions 
major minor  #blocks  name

   7        0      56780 loop0
   7        1      70848 loop1
   7        2      33048 loop2
 252        0   10485760 vda
 252        1   10372079 vda1
 252       14       4096 vda14
 252       15     108544 vda15
 252       16        366 vdb
 252       32    1048576 vdc
   8        0    1048576 sda
```

In this case `/dev/sda` is what we are looking for, and this is a disk provided by the iscsi target (running in the services VM) and shared by the iscsi initiators.

### DIFF
--- a/ha/virsh/delete-cluster.sh
+++ b/ha/virsh/delete-cluster.sh
@@ -2,9 +2,17 @@
 
 : "${UBUNTU_SERIES:=impish}"
 VM_PREFIX="fence-test-virsh-${UBUNTU_SERIES}-node"
+VM_SERVICES="services-${UBUNTU_SERIES}"
 
+# Remove all cluster nodes
 virsh list --state-running --name | grep "^$VM_PREFIX" | xargs -L 1 --no-run-if-empty virsh destroy
 virsh list --state-shutoff --name | grep "^$VM_PREFIX" | xargs -L 1 --no-run-if-empty virsh undefine --remove-all-storage
+
+# Remove VM running external services if it exists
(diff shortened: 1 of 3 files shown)


paride:
The changes LGTM, however I can't make the tests pass, so I'm clearly missing something :)

Console log: https://paste.ubuntu.com/p/9dqV3MK6F3/

I'm going to setup a reproducer.

lucaskanashiro:
I have tested in `dugtrio` with the latest changes and the tests are all passing now.

paride (approved):
LGTM and testsd fine, thanks!

PR #122, opened 2021-07-02
HA: fix curl calls in systemd resource agent tests

lucaskanashiro:
Those changes were tested in `dugtrio` and also in my host running Focal, everything seems to be working fine for me.

### DIFF
--- a/ha/virsh/run_tests.sh
+++ b/ha/virsh/run_tests.sh
@@ -9,10 +9,10 @@ source /etc/profile.d/libvirt-uri.sh
 
 test_failed=0
 for file in $TESTS; do
-  AGENT=$(echo "$file" | grep -oP '(?<=/).+(?=\_)')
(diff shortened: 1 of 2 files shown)


ha/virsh/run_tests.sh
```diff
@@ -9,10 +9,10 @@ source /etc/profile.d/libvirt-uri.sh
 
 test_failed=0
 for file in $TESTS; do
-  AGENT=$(echo "$file" | grep -oP '(?<=/).+(?=\_)')
+  AGENT=$(echo "$file" | grep -oP '(?<=/).+(?=\_)' | tr _ -)
```
paride, line 12: Is this change relevant to this fix? More in general, why do we need it?

lucaskanashiro: `hostname` command is removing the `_` which makes the teste fail. If we use only `-` this is not a problem.

paride (approved):
LGTM and works fine, thanks!

PR #125, opened 2021-07-14
HA: add an extra network interface to all VMs

lucaskanashiro:
This is a preparation to test the multipath fence agent. To have a minimal multipath setup we need at least 2 network interfaces in each node so that we can reach the target via different paths.

With that said, I added changes to create a new network using virsh and added it to all VMs. The only thing that I did not figure out how to fix it properly is to get an IP address in this second network interface, at the moment I am calling `dhclient` to assign an IP address to it, if you have a better solution for that, I'll be happy to incorporate to this PR.

And I also ran all the tests locally using this branch and all of them passed.

### DIFF
--- a/ha/virsh/create-vm.sh
+++ b/ha/virsh/create-vm.sh
@@ -18,6 +18,8 @@ PUB_KEY=$(cat "${PUB_KEY_FILE}")
 RAM=1024
 VCPU=1
 
+HA_NETWORK="ha"
+
 setup_workdir() {
   mkdir -p "${IMAGES_DIR}"/base \
 	   "${IMAGES_DIR}"/"${VM_NAME}" \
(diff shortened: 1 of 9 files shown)


paride (changes requested):
Hi, I can follow the logic of this and it looks good, however these changes cause test failures, see for example https://paste.ubuntu.com/p/fV9Dx2f8hs/ (coming from dugtrio).

lucaskanashiro:
@paride I tested the latest changes in `dugtrio` and all tests passed. Could you take a look?

lucaskanashiro:
I triggered this job and all tests passed as well: https://jenkins.ubuntu.com/server/view/ha/job/ha-agent-virsh-impish-all/4/

paride (approved):
LGTM

PR #127, opened 2021-08-03
OCI: Check whether the manifest file(s) is(are) empty

sergiodj:
Athos has recently worked on (and fixed) a bug on the golang manifest
generator script which rendered empty manifest files for some images.
Apparently this bug has silently crept in some time ago, and we didn't
notice.  I thought that it'd be useful to expand our current
"is-the-manifest-file-present" unit test and make it also error if the
manifest file is empty (which should never happen).

I wasn't entirely sure whether this should be a new test or not, but I
think it makes sense (and is cleaner) to do it in the existing test.

### DIFF
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -72,6 +72,7 @@ check_manifest_exists()
 {
     local id="${1}"
     local manifest_dir="/usr/share/rocks"
+    local found=0
 
     # The expected files for each image type.
     local possible_manifest_files="dpkg.query manifest.yaml snapcraft.yaml upstream"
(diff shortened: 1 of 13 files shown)


oci-unit-tests/helper/test_helper.sh
```diff
@@ -84,18 +85,35 @@ check_manifest_exists()
...
+
+    debug "Verifying that the manifest file(s) is(are) not empty"
+    for file in ${files}; do
+        fsize=$(docker exec "${id}" stat -c '%s' "${manifest_dir}/${file}" 2> /dev/null)
+        if [ "${fsize}" -eq 0 ]; then
+            echo "E: Manifest file ${file} is empty" > /dev/stderr
+            debug "file ${file} is empty"
+            return 1
+        else
+            debug "file ${file} has size ${fsize}"
```
athos-ribeiro, line 112: should we include the unit (bytes) after the size value here?

athos-ribeiro (approved):
Thanks, Sergio!

LGTM. I added a small nitpick inline. Go ahead an merge after (or before, in case you disagree) addressing it.

oci-unit-tests/helper/test_helper.sh
sergiodj: Ah, good point.  I will include it.

sergiodj:
> Thanks, Sergio!
> 
> LGTM. I added a small nitpick inline. Go ahead an merge after (or before, in case you disagree) addressing it.

Thanks for the review, Athos.  I agree with your comment and will update the test accordingly.

PR #128, opened 2021-08-05
OCI: Fix and improve the lma-integration-tests

sergiodj:
As I mentioned during standup, lma-integration-tests is very outdated
and a bit broken.  This series of commits aim at bringing these tests
to the present and also fix the problems I found during the process.

The commits are not hard to understand, and their messages should be
descriptive.  Let me know if you have any questions.

BTW, the shell scripts are not shellcheck-compliant but I didn't feel
motivated to fix this, so I'm leaving them as-is for now.


lma-integration/tests/Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM ubuntu:groovy
+FROM ubuntu:impish
```
paride, line 1: I fear that we're going to forget to update this. Do you think we could use `ubuntu:devel` here?

paride (commented):
Thanks Sergio, this LGTM but I have one inline question.

lma-integration/run-tests
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+set -e
+
+progdir=$(dirname "$(readlink -f "${0}")")
+
+# We have to build the image of the test container first.
+cd "${progdir}/tests/"
+docker build . -t localtest/lma-integration-tests
+cd -
```
athos-ribeiro, line 10: Could we just change the build context instead to read `docker build ${progdir}/tests/ -t localtest/lma-integration-tests`?

athos-ribeiro (approved):
LGTM;

I did leave a small nitpick inline and I +1 Paride's comment on the base image tags

lma-integration/tests/Dockerfile
sergiodj: Ah, good point.  I've adjusted the Dockerfile.

lma-integration/run-tests
sergiodj: I wasn't sure whether I could pass the build directory directly to `docker build` or not (because we use a `COPY . .` statement inside the Dockerfile, so I was afraid that it was going to affected by not chdir'ing).  Thanks for the tip; I've tested and verified that it indeed works :-).

sergiodj:
Thanks for the reviews, @paride and @athos-ribeiro.  I force-pushed the branch with the requested changes.

paride (approved):
LGTM, thanks!

athos-ribeiro:
Thanks, Sergio. LGTM

sergiodj:
Thanks, guys.  I'm merging this.

PR #129, opened 2021-08-09
oci cassandra: do not `docker run` in interactive mode

paride:
Do not `docker run` in interactive mode (drop the `-i`) but do allocate
a tty (`--tty`) so we can get stdout from the run command.


athos-ribeiro (approved):
Thanks, Paride :)

PR #131, opened 2021-08-26
OCI: Consolidate `$ROOTDIR` usage (and fix a few pending issues)

sergiodj:
I was testing the new telegraf OCI image for 21.10 from inside the
image's repository (i.e., I was invoking the unittest script while
outside its working directory), and noticed that the test was failing
because it was using `$PWD` indiscriminately.

I took the opportunity to improve the handling of these scenarios.  We
already had a `$ROOTDIR` variable being declared in the test_helper.sh
script, but it contained the wrong value and wasn't being used by
anyone.  I fixed it and converted the existing calls to `$(dirname
"$0")` and made them use `$ROOTDIR` instead.  While at it, I have also
removed the needless `$PATH` manipulation that was being done in the
same script.

### DIFF
--- a/oci-unit-tests/apache2_test.sh
+++ b/oci-unit-tests/apache2_test.sh
@@ -70,7 +70,7 @@ test_default_config_ipv6() {
 
 test_static_content() {
     debug "Creating all-defaults apache2 container"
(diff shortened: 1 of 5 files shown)


athos-ribeiro (approved):
LGTM!

Thanks, Sergio!

sergiodj:
Thanks, Athos.

PR #132, opened 2021-08-26
OCI: Update cortex unittest to reflect changes in the last upstream version

sergiodj:
Upstream has changed the log output of cortex's table manager
component, so we have to adjust the existing unittest to stop
expecting table-related strings in the log.

This has been tested with the upcoming 21.10 cortex OCI image as well
as with the existing images from the ubuntu namespace.


athos-ribeiro (approved):
LGTM! Tested with current images.

Thanks, Sergio

sergiodj:
Thanks, Athos.

PR #134, opened 2021-09-02
OCI: Add squid tests

athos-ribeiro:
This is the initial state of the new squid OCI tests. The first configuration file is just the concatenation of the default Ubuntu configuration files with the docs stripped out. The second, is just a minimal configuration file to ensure we can inject configuration into the image.

### DIFF
--- a/oci-unit-tests/squid_test.sh
+++ b/oci-unit-tests/squid_test.sh
@@ -1,0 +1,186 @@
+# shellcheck shell=dash
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+. "$(dirname "$0")/helper/common_vars.sh"
+
+# cheat sheet:
+#  assertTrue $?
+#  assertEquals ["explanation"] 1 2
+#  oneTimeSetUp()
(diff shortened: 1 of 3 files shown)


athos-ribeiro:
This launchpad MP is needed to properly test these changes: https://code.launchpad.net/~canonical-server/ubuntu-docker-images/+git/squid/+merge/408039

sergiodj (changes requested):
Thanks for the PR, Athos.

I like the tests!  I'm just leaving a few minor comments, but otherwise it LGTM.

For completeness: I was trying to run this using an older version of docker.io (`20.10.2-0ubuntu1~20.04.2`), and was seeing failures with the IPv6 test.  When I updated docker.io to the latest version available on Focal (`20.10.7-0ubuntu1~20.04.1`), the problem disappeared.

athos-ribeiro:
Thanks for the review, Sergio!

I addressed all your comments.

Yes, I agree with the approach of pulling the python image before the tests and removing it afterwards!

sergiodj:
Awesome, thanks Athos.  +1

PR #136, opened 2021-09-10
Bind9 tests

bryceharrington:
This adds some basic test cases for the new impish bind9 OCI image.  It does a local check that the service is present on the expected port, and a client/server test to check a trivial dns lookup.  This provides a similar level of coverage to what is available in the bind9 DEP8 tests.

Note that due to LP: #1943049 the bind9 image can't be built for impish, so I've been testing against a hirsute-built image.

Testing is performed via:

    DOCKER_IMAGE=test/bind9:edge sh ./bind9_test.sh

I also propose making a couple functions as helpers.  These are broken out as separate follow-on commits that can easily dropped or moved to other branches if desired.

### DIFF
--- a/oci-unit-tests/bind9_test.sh
+++ b/oci-unit-tests/bind9_test.sh
@@ -1,0 +1,104 @@
+# shellcheck shell=dash
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+. "$(dirname "$0")/helper/common_vars.sh"
+
+# cheat sheet:
+#  assertTrue $?
+#  assertEquals ["explanation"] 1 2
+#  oneTimeSetUp()
+#  oneTimeTearDown()
+#  setUp() - run before each test
+#  tearDown() - run after each test
+
+oneTimeSetUp() {
+    # Remove image before test.
+    remove_current_image
+
+    # Make sure we're using the latest OCI image.
+    docker pull --quiet "${DOCKER_IMAGE}" > /dev/null 2>&1
+
+    # Cleanup stale resources
+    tearDown
+    oneTimeTearDown
+
+    # Setup network
+    docker network create "${DOCKER_NETWORK}" > /dev/null 2>&1
+}
+
+oneTimeTearDown() {
+    docker network rm "${DOCKER_NETWORK}" > /dev/null 2>&1
+}
+
+tearDown() {
+    local container
+
+    for container in $(docker ps --filter "name=${DOCKER_PREFIX}" --format "{{.Names}}"); do
+        debug "Removing container ${container}"
+        stop_container_sync "${container}"
+    done
+}
+
+wait_bind9_container_ready() {
+    local container="${1}"
+    local log="running"
+    wait_container_ready "${container}" "${log}"
+}
+
+
+test_local_connection() {
+    local container
+
+    debug "Creating all-defaults bind9 container"
+    container=$(run_container_service)
+    assertNotNull "Failed to start the container" "${container}" || return 1
+
+    wait_bind9_container_ready "${container}" || return 1
+
+    debug "Installing test dependencies locally"
+    install_container_packages "${container}" "lsof" || return 1
+
+    debug "Checking for service on local port 53"
+    docker exec "${container}" lsof -i:53 | grep -q LISTEN
+
+    assertTrue ${?}
+}
+
+test_network_connection() {
+    local container_server
+    local container_client
+
+    debug "Creating all-defaults bind9 container (server)"
+    container_server=$(SUFFIX=server run_container_service)
+    assertNotNull "Failed to start the container" "${container_server}" || return 1
+
+    # Wait for server to be ready
+    wait_bind9_container_ready "${container_server}" || return 1
+
+    debug "Creating all-defaults bind9 container (client)"
+    container_client=$(SUFFIX=client run_container_service)
+    assertNotNull "Failed to start the container" "${container_client}" || return 1
+
+    # Wait for client to be ready
+    wait_bind9_container_ready "${container_client}" || return 1
+
+    debug "Installing test dependencies (client)"
+    install_container_packages "${container_client}" "bind9-dnsutils" || return 1
+
+    # Use dig from the client to the server's IP
+    docker exec "${container_client}" dig "${DOCKER_PREFIX}_server:30053" > /dev/null
+    assertTrue ${?}
+}
+
+test_manifest_exists() {
+    local container
+
+    debug "Testing that the manifest file is available in the image"
+    container=$(run_container_service)
+
+    check_manifest_exists "${container}"
+    assertTrue "Manifest file(s) do(es) not exist in image" ${?}
+}
+
+load_shunit2
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -24,6 +24,42 @@ debug() {
     fi
 }
 
+# Creates a new docker container from $DOCKER_IMAGE.
+#
+# By default, creates a temporary container running the software under
+# test as a daemon service.  The container is automatically removed
+# when the daemon exits.  All parameters to run_container_service() are
+# passed directly through to the contained software daemon.  The docker
+# behavior can be adjusted through environmental variables listed
+# below.  In particular, $suffix can be used to distinguish between
+# multiple containers used in a given test, such as 'client' and
+# 'server'.
+#
+# Parameters:
+#   $@: parameters to pass through to the container
+# Environment:
+#   $DOCKER_IMAGE: Name of the image containing the software to be tested.
+#       Defaults to "docker.io/ubuntu/<package>:edge"
+#   $DOCKER_PREFIX: Common prefix for all containers for this test.
+#       Defaults to "oci_<package>_test".
+#   $DOCKER_NETWORK: User-created network to connect the container to.
+#       Defaults to "oci_<package>_test_net".
+#   $SUFFIX: Optional tag for ensuring unique container names.
+#       Defaults to a random string.
+#
+# Stdout: Name of created container.
+# Returns: Error code from docker, or 0 on success.
+run_container_service() {
+    SUFFIX=${SUFFIX:-$(tr -dc 'a-zA-Z0-9' < /dev/urandom | fold -w 8 | head -n 1)}
+    docker run \
+       --network "${DOCKER_NETWORK}" \
+       --rm \
+       -d \
+       --name "${DOCKER_PREFIX}_${SUFFIX}" \
+       "${DOCKER_IMAGE}" \
+       "$@"
+}
+
 # $1: container id
 # $2: timeout (optional).  If not specified, defaults to 10 seconds
 stop_container_sync() {
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -53,15 +53,15 @@ wait_container_ready() {
 
     debug -n "Waiting for container to be ready "
     while ! docker logs "${id}" 2>&1 | grep -qE "${msg}"; do
-	sleep 1
-	timeout=$((timeout - 1))
-	if [ $timeout -le 0 ]; then
-	    fail "ERROR, failed to start container ${id} in ${max} seconds"
-	    echo "Current container list (docker ps):"
-	    docker ps
-	    return 1
-	fi
-	debug -n "."
+        sleep 1
+        timeout=$((timeout - 1))
+        if [ $timeout -le 0 ]; then
+            fail "ERROR, failed to start container ${id} in ${max} seconds"
+            echo "Current container list (docker ps):"
+            docker ps
+            return 1
+        fi
+        debug -n "."
     done
     debug "done"
 }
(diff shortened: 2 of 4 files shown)


paride:
Hi @bryceharrington, in general this LGTM, some comments/questions:
1. Is there an easy way for me to pull the `test/bind9` to actually try this out?
2. There's another occurrence of `apt-get install` in `memcached_test.sh` you may want to replace with `install_container_packages`
3. This PR introduces a shellcheck warning (missing quotes):
```
In test_helper.sh line 114:
    for package in ${@}; do
                   ^--^ SC2068: Double quote array expansions to avoid re-splitting elements.
```
4. I see you converted some indentation from spaces to tabs. We aren't following a consistent style in these test scripts, there is a strange mix where 1st level indent is done with spaces and deeper levels with tabs. I think this is the result of some editor auto-coverting 7-spaces to a tab or something like that. We should switch to a consistent style, and that's what you were probably trying to move forward to, but I don't see exactly what you were aiming at...
5. I take that switching the other tests to `run_container_server()` is out of scope for this PR, but should we leave at least a note to ourselves (a GH issue or a TODO file) mentioning that we should switch to the new helper? I fear that we'll keep using the old approach when writing new tests by taking a random one as a template.

oci-unit-tests/bind9_test.sh
```diff
@@ -0,0 +1,89 @@
...
+	debug "Removing container ${container}"
+	stop_container_sync "${container}"
+    done
+}
+
+test_local_connection() {
+    local container
+
+    debug "Creating all-defaults bind9 container"
+    container=$(run_container_server)
```
sergiodj, line 47: After starting the container, it is a good practice to wait for it to be ready before executing the commands.  Take a look at the `wait_.*_container_ready` family of functions for some examples.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -66,6 +102,31 @@ wait_container_ready() {
...
+{
+    local retval=0
+    local id="${1}"
+    shift
+
+    docker exec -u root "${id}" apt-get -qy update > /dev/null
+    for package in ${@}; do
+	debug "Installing ${package} into ${id}"
+	if docker exec -u root "${id}" \
+	       apt-get -qy --option=Dpkg::options::=--force-unsafe-io install "${package}" \
```
sergiodj, line 117: I was wondering about the `--option=Dpkg::options::=--force-unsafe-io` option, but then I saw that it came from the `apt-get` command present in the memcached unit test.

I understand the rationale behind it, but I'm wondering if it's really needed.  I think it causes some confusion and you can just get rid of it here IMO.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -15,15 +15,51 @@ load_shunit2() {
...
-            echo -n "$@"
-        else
-            echo "$@"
-        fi
+	if [ "${1}" = "-n" ]; then
+	    shift
+	    echo -n "$@"
+	else
+	    echo "$@"
+	fi
```
sergiodj, line 23: Are these re-indentations intentional?  I don't like the idea of mixing things in the same PR; this should be about bind9.  It's OK if you want to propose a new indentation for this file, but I think another PR should be filed for that.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -15,15 +15,51 @@ load_shunit2() {
...
+#   $DOCKER_PREFIX: Common prefix for all containers for this test.
+#       Defaults to "oci_<package>_test".
+#   $DOCKER_NETWORK: User-created network to connect the container to.
+#       Defaults to "oci_<package>_test_net".
+#   $SUFFIX: Optional tag for ensuring unique container names.
+#       Defaults to a random string.
+#
+# Stdout: Name of created container.
+# Returns: Error code from docker, or 0 on success.
+run_container_server() {
```
sergiodj, line 52: I think many of our existing tests can benefit from this generic function (not all of them, though).  WDYT about search-and-replacing this function in our tests?  I think doing this in another PR would be better.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -15,15 +15,51 @@ load_shunit2() {
...
+# Creates a new docker container from $DOCKER_IMAGE.
+#
+# By default, creates a temporary container running the software under
+# test as a daemon service.  The container is automatically removed
+# when the daemon exits.  All parameters to run_container_server() are
+# passed directly through to the contained software daemon.  The docker
+# behavior can be adjusted through environmental variables listed
+# below.  In particular, $suffix can be used to distinguish between
+# multiple containers used in a given test, such as 'client' and
+# 'server'.
```
sergiodj, line 36: Because containers can be either servers of clients, WDYT about renaming this function to something more generic like `run_container_application` or `run_container_service`?

sergiodj (changes requested):
Thanks for the PR, @bryceharrington.

I'm leaving some comments about things that I think will need to be fixed/adjusted.  @paride has also made some good points.

sergiodj:
>     1. Is there an easy way for me to pull the `test/bind9` to actually try this out?

I can answer this one :-).  The `test/bind9` image needs to be built locally.  Because of the dreaded glibc + docker.io bug, you won't be able to build the impish image.  For this reason, you will need to `s/impish/hirsute/` in the Dockerfile.

bryceharrington:
Thanks for the review paride

> Hi @bryceharrington, in general this LGTM, some comments/questions:
>     2. There's another occurrence of `apt-get install` in `memcached_test.sh` you may want to replace with `install_container_packages`
Fixed
>     3. This PR introduces a shellcheck warning (missing quotes):
Fixed
>     1. I see you converted some indentation from spaces to tabs. We aren't following a consistent style in these test scripts, there is a strange mix where 1st level indent is done with spaces and deeper levels with tabs. I think this is the result of some editor auto-coverting 7-spaces to a tab or something like that. We should switch to a consistent style, and that's what you were probably trying to move forward to, but I don't see exactly what you were aiming at...

Not intentional; by default emacs tries to be "helpful" but sometimes makes wrong guesses.  I usually override the tab/space settings on a per-project basis but just hadn't done that yet on oci-unit-tests.  I switched the new routines to be all space indents.

>     2. I take that switching the other tests to `run_container_server()` is out of scope for this PR, but should we leave at least a note to ourselves (a GH issue or a TODO file) mentioning that we should switch to the new helper? I fear that we'll keep using the old approach when writing new tests by taking a random one as a template.

I was going to offer to fix up the rest of the tests as follow up PR(s) if this looked acceptable.  I also thought it'd be out of scope for this PR, plus could lead to merge conflicts.

oci-unit-tests/bind9_test.sh
bryceharrington: Thanks, fixed.

oci-unit-tests/helper/test_helper.sh
bryceharrington: It seems to be some sort of optimization, and I assumed that was necessary for docker containers for some reason I just wasn't aware of.  It is indeed adding confusion, I'll drop it.  I suppose if we add optimizations like that we should make sure to document them.

oci-unit-tests/helper/test_helper.sh
bryceharrington: Not intentional, I habitually run a whitespace cleanup routine in my editor and I guess it had its own ideas about how things should be indented...  I've backed these changes out.

oci-unit-tests/helper/test_helper.sh
bryceharrington: Completely agree on both points, and will try to send a PR next week.

oci-unit-tests/helper/test_helper.sh
bryceharrington: Yes, those both sound better.  I think since this routine specifically runs daemon services (ala the -d flag), "run_container_service" is probably the more intuitive name.  I'll rename.

bryceharrington:
I've pushed the resulting changes, and hopefully the whitespace looks better.

sergiodj:
> Not intentional; by default emacs tries to be "helpful" but sometimes makes wrong guesses. I usually override the tab/space settings on a per-project basis but just hadn't done that yet on oci-unit-tests. I switched the new routines to be all space indents.

Ah, I'm an Emacs user myself and suffered from the same problem in the beginning.  I thought about contributing a `.dir-locals.el` but I'm not sure if it'd be "too much" (given that everybody else in the team seems to be a vim user).

bryceharrington:
Do you use the same emacs settings for server-test-scripts as you do for the oci image repos themselves?  I see the way I've set up .dir-locals.el allows tabs generally but disables them case-by-case.

What I use is just:
((nil . ((indent-tabs-mode . nil))
))

Do you have more than this in yours?

sergiodj (changes requested):
There are a few minor things that need fixing (one of the comments is more a personal preference than an issue, so feel free to ignore).

I tried to run the test locally (using `hirsute` when building the image), and I'm seeing a failure like this:

```
$ DOCKER_IMAGE=sergiodj/bind9 sh bind9_test.sh 
test_local_connection
named     1 bind   73u  IPv4 154392660      0t0  TCP localhost:domain (LISTEN)
named     1 bind   83u  IPv4 154389885      0t0  TCP 48f5a8679cba:domain (LISTEN)
test_network_connection
ASSERT:ERROR, failed to start container  in 60 seconds
Current container list (docker ps):
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
test_manifest_exists

Ran 3 tests.

FAILED (failures=1)
```

I haven't investigated it further, but I'm wondering if you can reproduce it.

Thanks!

sergiodj:
> Do you use the same emacs settings for server-test-scripts as you do for the oci image repos themselves? I see the way I've set up .dir-locals.el allows tabs generally but disables them case-by-case.
> 
> What I use is just:
> ((nil . ((indent-tabs-mode . nil))
> ))
> 
> Do you have more than this in yours?

After having many episodes where my Emacs would auto-reindent a lot of things when hacking these shell scripts, I've added the following to my `init.el`:

```elisp
(add-hook 'sh-mode-hook (lambda () (setq indent-tabs-mode nil)))
```

So for me this is a global configuration now, not per-project.

oci-unit-tests/bind9_test.sh
```diff
@@ -0,0 +1,99 @@
...
+}
+
+test_local_connection() {
+    local container
+
+    debug "Creating all-defaults bind9 container"
+    container=$(run_container_service)
+    assertNotNull "Failed to start the container" "${container}" || return 1
+
+    wait_bind9_container_ready "${container}"                    || return 1
```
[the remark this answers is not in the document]
bryceharrington: ok

oci-unit-tests/bind9_test.sh
```diff
@@ -0,0 +1,99 @@
...
+
+    debug "Installing test dependencies locally"
+    install_container_packages "${container}" "lsof"             || return 1
+
+    debug "Checking for service on local port 53"
+    docker exec "${container}" lsof -i:53 | grep LISTEN
```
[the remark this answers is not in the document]
bryceharrington: Added.  I'd left that out to debug but agree it looks cleaner with -q.

oci-unit-tests/bind9_test.sh
```diff
@@ -0,0 +1,99 @@
...
+
+    for container in $(docker ps --filter "name=${DOCKER_PREFIX}" --format "{{.Names}}"); do
+	debug "Removing container ${container}"
+	stop_container_sync "${container}"
+    done
+}
+
+wait_bind9_container_ready() {
+    local container="${1}"
+    local log="Ready to accept connections"
```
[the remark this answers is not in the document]
bryceharrington: Yep this was one of the causes of the test failures, thanks.

bryceharrington:
I've added in the suggestions, and a couple other fixups I discovered in re-running the tests related to the new wait-for-server functionality.  The test cases are passing again for me.

I've also reorganized the commits that should make things cleaner and clearer for future commit history reviewers, by moving the two new helpers to discrete commits preceding the bind9 test, and squashed all the review fixes for the bind9 test into that commit.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -79,19 +140,19 @@ check_manifest_exists()
...
-	debug "not found"
-	return 1
+        debug "not found"
+        return 1
     fi
     debug "done"
 
     debug "Verifying whether the manifest file(s) exist"
     for want_file in ${possible_manifest_files}; do
-	for got_file in ${files}; do
```
sergiodj, line 89: I have the impression that you had removed these indentation changes during the last iteration, but now I see them back here.  Or maybe I'm misremembering?

sergiodj (approved):
Thanks, LGTM.

I'm approving this PR, but maybe @paride would like to take a second look at it so I think it's safer to wait until Monday to merge it.

On a side note, I'm leaving a comment about the indentation changes.  I could swear they were not present in the last iteration, but I'm not 100% certain.  Either way, it's a minor thing.

Thanks.

oci-unit-tests/helper/test_helper.sh
bryceharrington: Yeah I included a new whitespace fixup commit that converts the tabs to spaces per our discussion to make the file's whitespace consistent.

That patch can be dropped without any conflict if you prefer keeping the tabs.

sergiodj: No, it's alright.  I just thought that it'd be better to open a new PR to reindent everything, but I have no problem doing it here.  Thanks.

paride (approved):
This LGTM and the tests pass fine using a locally built image based on Hirsute (thanks Sergio).

PR #139, opened 2021-11-22
oci: sleep 5s on wait_container_ready()

paride:
nginx is not the only image requiring some extra sleep time to make sure
the container is actually ready: squid is another example. Let's move
the sleep in the wait_container_ready() helper function; all the images
will get some extra time, but it won't hurt.

### DIFF
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -99,6 +99,7 @@ wait_container_ready() {
         fi
         debug -n "."
     done
+    sleep 5
     debug "done"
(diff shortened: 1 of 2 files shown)


athos-ribeiro (approved):
LGTM!

Although, this is also reducing the nginx sleep. Hopefully 5 seconds won't make the test flaky

paride:
@athos-ribeiro yeah I tested it (locally) and it *seems* to be enough for nginx, 10s were probably way more than enough. We can still bump it if the test starts being flaky. Thanks for reviewing!

PR #140, opened 2021-12-02
ha lvm-activate: bump sleep in test_move_resource

paride:
The test is racey with a 1s sleep; seems stable with 3s.


paride:
I did some manual testing and this seems to fix the `ha-agent-virsh-impish-resource_lvm-activate` flaky job.

lucaskanashiro:
Thanks @paride , LGTM. I want to find the time to work more on this but likely next year only.

PR #141, opened 2021-12-06
copr: switch to the copr api v3

paride:
The v2 api is broken at the moment [1], and even if fixed it's going to
be turned off due to deprecation in a few months. Let's switch to v3.

[1] https://pagure.io/copr/copr/issue/2003
[2] https://fedora-copr.github.io/posts/EOL-APIv1-APIv2-pt2

### DIFF
--- a/cloud-init/copr_build.py
+++ b/cloud-init/copr_build.py
@@ -12,7 +12,7 @@ import os
 import sys
 import time
 
-import copr
+from copr.v3 import Client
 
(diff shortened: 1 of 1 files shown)


paride:
Tested on torkoal, works fine.

This is not super nice as it's more an adaptation of our api v2 script to the v3 api, without really rethinking it with the new logic in mind, but it works.

TheRealFalcon (approved):
Thanks!

PR #142, opened 2021-12-10
oci: Verify required manifest files presence

athos-ribeiro:
This will break tests for the following images, which need fixing:

- cassandra
- cortex
- grafana
- prometheus
- prometheus-alertmanager

### DIFF
--- a/oci-unit-tests/helper/test_helper.sh
+++ b/oci-unit-tests/helper/test_helper.sh
@@ -129,14 +129,21 @@ install_container_packages()
 }
 
 # $1: container id
+# ${2...}: required manifest files
 check_manifest_exists()
 {
     local id="${1}"
+    shift
     local manifest_dir="/usr/share/rocks"
-    local found=0
+    local found=""
 
     # The expected files for each image type.
     local possible_manifest_files="dpkg.query manifest.yaml snapcraft.yaml upstream"
+    if [ -n "${1}" ]; then
+        local required_manifest_files="dpkg.query ${*}"
+    else
+        local required_manifest_files="dpkg.query"
+    fi
 
     debug -n "Listing the manifest file(s) inside ${id}"
     local files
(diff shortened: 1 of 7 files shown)


oci-unit-tests/helper/test_helper.sh
```diff
@@ -151,15 +158,28 @@ check_manifest_exists()
         for got_file in ${files}; do
             if [ "${got_file}" = "${want_file}" ]; then
                 debug "found ${got_file}"
-                found=1
+                if [ -n "${found}" ]; then
```
sergiodj, line 161: `${found}` is initialized with 0, so it will never be empty.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -151,15 +158,28 @@ check_manifest_exists()
...
 
-    if [ "${found}" -eq 0 ]; then
+    if [ -z "${found}" ]; then
         echo "E: No manifest file found in the image" > /dev/stderr
         debug "not found"
         return 1
+    else
+      debug "Verifying required manifest file(s) presence [${required_manifest_files}]"
+        for required_file in ${required_manifest_files}; do
+            if ! echo "${found}" | grep -q -w "${required_file}"; then
```
sergiodj, line 177: I'd prefer `grep -qFw` just because we know `${required_file}` is never going to be a regexp.

sergiodj (changes requested):
Thanks for the PR, Athos.

There are a few nits, but overall it LGTM and I agree with the proposed change.  Do you prefer to submit another PR to adjust the `*_test.sh` files, or do you think it's fine to include everything in this one?

athos-ribeiro:
> Thanks for the PR, Athos.
> 
> There are a few nits, but overall it LGTM and I agree with the proposed change. Do you prefer to submit another PR to adjust the `*_test.sh` files, or do you think it's fine to include everything in this one?


Hi @sergiodj,

Thanks for the review!

I addressed your comments (nice catch BTW, thanks).

I will file a follow up PR to adjust the individual test files if that is OK :)

athos-ribeiro:
On a second thought, this is too simple for a second PR. I included an additional commit with those changes.

oci-unit-tests/helper/test_helper.sh
```diff
@@ -129,14 +129,20 @@ install_container_packages()
 }
 
 # $1: container id
+# ${2...}: required manifest files
 check_manifest_exists()
 {
     local id="${1}"
+    shift
     local manifest_dir="/usr/share/rocks"
-    local found=0
```
sergiodj, line 136: Can you instead initialize `$found` with an empty value instead?  This way, we make sure it stays local when it's assigned down below.

sergiodj (approved):
Thanks for the quick follow up, Athos.

I'm approving this PR, but there's a small nit that needs to be fixed so I'm leaving a comment about it.

We've discussed this in private already, but the idea here is to fix all images first before merging this PR in order to prevent known failures to happen in our CI.

Thanks.

oci-unit-tests/helper/test_helper.sh
athos-ribeiro: Thanks, Sergio.

Sure! I just pushed the change [1].

[1] https://github.com/canonical/server-test-scripts/compare/b48487295c744c9394112dcb524eebb7bfb2c1f2..caf165ebe436d6ac02eb723cdb343c02a1550842

sergiodj: Awesome, thanks.

PR #143, opened 2021-12-14
oci: prometheus testing improvements

athos-ribeiro:
This set of changes ensures prometheus tests are fit to run against all currently published versions of prometheus by considering recent UI changes and output redirections.


paride (approved):
LGTM, thanks!

PR #144, opened 2022-01-18
oci: build cqlsh image from a release tag

athos-ribeiro:
cqlsh is used to test the published Cassandra images. We currently do so
with the development version of cqlsh. This patch changes this behavior
to use a release version of cqlsh to avoid development issues with the
project builds.

Signed-off-by: Athos Ribeiro <athos.ribeiro@canonical.com>

### DIFF
--- a/oci-unit-tests/cassandra_test_data/Dockerfile
+++ b/oci-unit-tests/cassandra_test_data/Dockerfile
@@ -1,8 +1,9 @@
 FROM ubuntu:impish
+ARG CQLSH_VERSION=4.0.1
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get full-upgrade -y && \
 	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
 		git python3-pip build-essential python3-dev && \
-	git clone https://github.com/apache/cassandra.git && \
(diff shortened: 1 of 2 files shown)


oci-unit-tests/cassandra_test_data/Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM ubuntu:impish
+ARG CQLSH_VERSION=4.0.1
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get full-upgrade -y && \
 	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
 		git python3-pip build-essential python3-dev && \
-	git clone https://github.com/apache/cassandra.git && \
+	git clone -b cassandra-${CQLSH_VERSION} --single-branch https://github.com/apache/cassandra.git && \
```
sergiodj, line 6: WDYT about using `--depth 1` (which implicitly means `--single-branch`) here to create a shallow clone (and save bandwidth/time)?

sergiodj (approved):
Thanks, LGTM.  The script runs fine locally.  I'm leaving just a suggestion re. the `git clone` command; feel free to adopt it if you think it's useful.  Otherwise, +1.

oci-unit-tests/cassandra_test_data/Dockerfile
athos-ribeiro: +1! Fixed

PR #145, opened 2022-02-01
oci: fix dig call in bind9 network tests

athos-ribeiro:
* LP: #1955680

The dig call should perform some query (it should return 0 even if the query cannot be resolved) in the bind9 server.


paride (approved):
LGTM, thanks!

PR #146, opened 2022-02-02
Fix bind9 local port tests

athos-ribeiro:
* Check if bind9 accepts connections for both TCP and UDP
* Use ss instead of lsof to leverage installed package (faster tests)
* remove redundant default port from dig call


paride:
I'm doing a test run of this in Jenkins.

PR #147, opened 2022-02-09
oci: provide assertion error data

athos-ribeiro:
Some of our assertions will fail silently. These patches fix the issue for bind9 and grafana


paride (approved):
LGTM, thanks!

PR #148, opened 2022-02-10
c-i copr_test: run tests on CentOS Streams

paride:
Switching to Streams requires enabling EPEL-Next, which contains
EPEL packages rebuilt on Streams, see [1].

[1] https://docs.fedoraproject.org/en-US/epel/

### DIFF
--- a/cloud-init/copr_test
+++ b/cloud-init/copr_test
@@ -149,7 +149,10 @@ main() {
     name="copr-centos-${uuid%%-*}"
     repo="copr.repo"
 
-    start_container "images:centos/$version" "$name"
+    local image="images:centos/$version"
+    # CentOS >= 8 only exists as CentOS Stream
+    ((version >= 8)) && image+="-Stream"
+    start_container "$image" "$name"
 
     tee "$TEMP_D/$repo" <<EOF >/dev/null
 [group_cloud-init-cloud-init-dev]
@@ -170,6 +173,12 @@ EOF
 
     install_pkgs "$name" epel-release ||
         fail "failed to install epel-release"
+
+    # CentOS Stream needs EPEL Next on top of EPEL
+    if ((version >= 8)); then
+        install_pkgs "$name" epel-next-release ||
+            fail "failed to install epel-next-release"
+    fi
 
     install_pkgs "$name" cloud-init ||
         fail "failed: installing cloud-init"


cloud-init/copr_test
```diff
@@ -149,7 +149,7 @@ main() {
     name="copr-centos-${uuid%%-*}"
     repo="copr.repo"
 
-    start_container "images:centos/$version" "$name"
+    start_container "images:centos/${version}-Stream" "$name"
```
holmanb, line 152: I don't think this works for our current tests, since Centos 7 is still supported (through mid-2024)  and as far as I can tell Centos Stream 7 doesn't exist. Maybe we could make the change: `s/version/centos_version/`, since it looks like our `centos_version` in server-jenkins-jobs matches the lxd image name.

cloud-init/copr_test
```diff
@@ -168,8 +168,8 @@ EOF
     lxc file push "$TEMP_D/$repo" "$name/etc/yum.repos.d/$repo" ||
         fail "failed: setting up copr.repo"
 
-    install_pkgs "$name" epel-release ||
-        fail "failed to install epel-release"
+    install_pkgs "$name" epel-release epel-next-release ||
```
holmanb, line 171: We might need to encode packages in `jenkins-jobs/cloud-init/centos.yaml` eventually, since the current Stream 9 installs directly from an http address[1]. Since `python3-httpretty` is not yet available for CentOS Stream 9, however, we can just "wait and see". Maybe that won't be needed when our dependencies are sorted out in EPEL 9.

[1] https://docs.fedoraproject.org/en-US/epel/#_el9

cloud-init/copr_test
paride: Right, thanks for spotting this. Thing is we *also* need the bare CentOS version later in the script, to grab the .rpm from the right COPR location. I think we need to infer when Stream is needed based on the requested version.

paride: I would have preferred if the LXD image kept the `centos/N` name for CentOS Stream, after all there's no CentOS non-Streams >= 8, and there's no CentOS Stream 7. I guess they went with the `N-Stream` name because for a period there were both CentOS 8 and CentOS Stream 8.

holmanb (approved):
Thanks for this @paride. LGTM!

PR #150, opened 2022-02-16
Test fence sbd

lucaskanashiro:
For the moment, I am installing `fence-agents-extra` during the test but it should be removed once `fence_sbd` is added to the `fence-agents-common` binary package.


paride (approved):
LGTM, tested manually, works fine. Thanks!

PR #152, opened 2022-02-23
Test iSCSI initiator resource

lucaskanashiro:


paride (approved):
Works fine both locally and in Jenkins, shellcheck clean, LGTM.

PR #154, opened 2022-03-16
Test fence_ipmilan

lucaskanashiro:
Add some basic tests for `fence_ipmilan`. To do that `ipmi_sim` (provided by `openipmi`) was used to simulate IPMI and do not need real hardware with that feature. In order to make it simple, its tests run in a single VM and not in a corosync/pacemaker cluster, calling `fence_ipmilan` directly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 results.xml
 ha/virsh/config/
 ha/virsh/images/
+ha/virsh/virtualbmc.venv/
--- a/ha/virsh/create-vm.sh
+++ b/ha/virsh/create-vm.sh
@@ -4,11 +4,11 @@ set -eux -o pipefail
 
 : "${UBUNTU_SERIES:=jammy}"
 
-VM_NAME="${1}"
-WORK_DIR="${2:-$(pwd)}"
-CONFIG_DIR="${3:-"${WORK_DIR}/config"}"
-IMAGES_DIR="${4:-"${WORK_DIR}/images"}"
-PUB_KEY_FILE="${5:-"/home/$(whoami)/.ssh/id_rsa.pub"}"
+VM_NAME="test"
+WORK_DIR=$(pwd)
+CONFIG_DIR="${WORK_DIR}/config"
+IMAGES_DIR="${WORK_DIR}/images"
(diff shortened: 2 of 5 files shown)


lucaskanashiro:
@paride these tests would make some changes in the host but at the end everything is undo, this is not ideal. If you have suggestions on how to improve it let me know.

lucaskanashiro:
@paride  today I tried to set up the `ipmi_sim` service in another VM to monitor the `ipmi-simulator` VM but it did not work. In the `lan.conf` file we have a `startcmd` which is supposed to be executed when the simulator is started or the monitored VM is turned on, and I did not find a way to make it work with those two VMs. I tried to make it work using `ssh` and calling `virsh` from the host machine but it did not work out.

Do we have nested virtualization available in the nodes running the jobs? If yes, we can run what we are already doing in a VM and spawn another one inside it.

I also looked for other options and I did find `vmbc` [1] as something viable, but unfortunately, it seems to not be available in the archive. I do not think installing it from PyPi is a good idea.

[1] https://docs.openstack.org/virtualbmc/latest/user/index.html

paride:
@lucaskanashiro nested virtualization should work.

On `vbmc`: if we can install it in a VM I think it's fine to install via `pip --user`. Otherwise we can install it on the host machine in a venv, something like:
```
python3 -m venv virtualbmc.venv
source virtualbmc.venv/bin/activate
pip install -U pip
pip install virtualbmc
```
Pip will cache the packages it downloads so it won't re-download them every time. We are already doing this for some jobs, where packaging is not available (e.g. for `azure-cli`). The [1] page you linked says

> or you can just run vbmcd from command line if you do not need the tool running persistently on the system. Once the server is up and running, you can use the vbmc tool to configure your libvirt domains as if they were physical hardware servers.

which looks like what we need. The only worry I have is that vbmc could interfere with the impi of the bare metal host machine, but looks like it only binds to libvirt, so hopefully it's safe.

lucaskanashiro:
@paride I think it is now better than before, let me know if you have other suggestions. The new implementation is using `virtualbmc`.

lucaskanashiro:
I addressed your comment @paride , ready for review again.

PR #157, opened 2022-04-08
HA: use fence_sbd instead of external/sbd in the test

lucaskanashiro:
I was mistakenly using `external/sbd` but that is not the agent that was
promoted to main, the correct one is `fence_sbd`.


athos-ribeiro (approved):
LGTM

PR #160, opened 2022-04-22
Update kafka tests to comply with new structure

athos-ribeiro:
The kafka image was split into a kafka and a zookeeper image. Here, we
adapt the tests to the new images structure.

Signed-off-by: Athos Ribeiro <athos.ribeiro@canonical.com>

### DIFF
--- a/oci-unit-tests/kafka_test.sh
+++ b/oci-unit-tests/kafka_test.sh
@@ -12,6 +12,7 @@
 #  setUp() - run before each test
 #  tearDown() - run after each test
 
+readonly ZOOKEEPER_IMAGE="${ZOOKEEPER_IMAGE:-${DOCKER_REGISTRY}/${DOCKER_NAMESPACE}/zookeeper:${DOCKER_TAG}}"
 readonly ZOOKEEPER_PORT=2181
 readonly BROKER_PORT=9092
 
@@ -53,8 +54,7 @@ docker_run_zookeeper() {
      -d \
      -p "${ZOOKEEPER_PORT}":"${ZOOKEEPER_PORT}" \
      --name "${DOCKER_PREFIX}"_zookeeper_"${id}" \
-     "${DOCKER_IMAGE}" \
-     zookeeper-server-start.sh /etc/kafka/zookeeper.properties
+     "${ZOOKEEPER_IMAGE}"
 }
 
 # Helper function to execute kafka with some common arguments.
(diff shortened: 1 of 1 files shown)


sergiodj:
Thanks, Athos.

Running the tests locally (using the images published on Dockerhub) gives me the following:

```
test_local_port_binding
test_communication
test_producer_consumer
ASSERT:Could not consume message. Was the message produced?
test_manifest_exists

Ran 4 tests.

FAILED (failures=1)
```

I remember running these tests before the split and they were passing, so maybe the error was introduced after the split?

Ah, a small nit: could you also prefix the commit message with `OCI: `, please?

oci-unit-tests/kafka_test.sh
```diff
@@ -53,8 +54,7 @@ docker_run_zookeeper() {
      -d \
      -p "${ZOOKEEPER_PORT}":"${ZOOKEEPER_PORT}" \
      --name "${DOCKER_PREFIX}"_zookeeper_"${id}" \
-     "${DOCKER_IMAGE}" \
-     zookeeper-server-start.sh /etc/kafka/zookeeper.properties
+     "${ZOOKEEPER_IMAGE}"
 }
 
 # Helper function to execute kafka with some common arguments.
```
sergiodj, line 60: This comment needs to be updated.

sergiodj (changes requested):
Aside from a small request to update a comment, everything else LGTM.  I'm marking this review as "Request changes" though, because one of the tests is failing locally.

athos-ribeiro:
Thanks, Sergio.

The tests are not failing here. Any change you were using the old images locally?

I updated the commit message as requested. I am not sure if I understood the request for updating the comment for running the kafka image though....

sergiodj:
> The tests are not failing here. Any change you were using the old images locally?

Sigh, it seems like it.  I removed the kafka and zookeeper images and ran the test again, and now it passes.  But that's strange; the `oneTimeSetUp` function is calling `remove_current_image`, which should have taken care of refreshing the kafka image for me (it doesn't refresh the zookeeper image, but I didn't have it the first time I ran the test anyway).  Hmm...

> I updated the commit message as requested. I am not sure if I understood the request for updating the comment for running the kafka image though....

Thanks.  The request to update the comment was based on a misunderstanding: I thought that `docker_run_broker` was running something else other than kafka, but it turns out that it's actually running kafka itself.  I'm OK with keeping this function name, but I would personally have named the function `docker_run_kafka`, I think.  Anyway, not a big deal.

Thanks, +1 from me.

sergiodj (approved):

PR #161, opened 2022-05-05
copr_test: new flags: --name, --vm (SC-989)

paride:
New flags:
  --name: allows to specify the container name from the command line
    --vm: use a LXD VM instead of a container.


paride:
Companion change: https://github.com/canonical/cloud-init/commit/d1149988323bb88ec8ee6aa14bdd9d0618582771

PR #162, opened 2022-05-16
ha: use a test-dependent ha network name

paride:
This way if tests run in parallel they won't reconfigure/delete
each other's network.


paride:
@lucaskanashiro I hope this will remove some occasional flakyness.

lucaskanashiro (approved):
Thanks for the proposed changes Paride. All looks good to me. I also did run some tests locally and everything worked as expected.

PR #163, opened 2022-05-16
oci kafka: bump timeout in "consuming test message"

paride:
We're hitting the 5000ms timeout in automation.


athos-ribeiro:
LGTM!

PR #164, opened 2022-08-12
HA: move to pcs

lucaskanashiro:
Update all HA related scripts to use `pcs` instead of `crmsh`.

### DIFF
--- a/ha/virsh/tests/fence_mpath_test.sh
+++ b/ha/virsh/tests/fence_mpath_test.sh
@@ -16,10 +16,10 @@ oneTimeSetUp() {
 }
 
 configure_cluster_properties() {
-  run_command_in_node "${IP_VM01}" "sudo crm configure property stonith-enabled=on"
-  run_command_in_node "${IP_VM01}" "sudo crm configure property stonith-action=reboot"
-  run_command_in_node "${IP_VM01}" "sudo crm configure property no-quorum-policy=stop"
-  run_command_in_node "${IP_VM01}" "sudo crm configure property have-watchdog=false"
+  run_command_in_node "${IP_VM01}" "sudo pcs property set stonith-enabled=on"
+  run_command_in_node "${IP_VM01}" "sudo pcs property set stonith-action=reboot"
+  run_command_in_node "${IP_VM01}" "sudo pcs property set no-quorum-policy=stop"
+  run_command_in_node "${IP_VM01}" "sudo pcs property set have-watchdog=false"
 }
 
 add_res_key_to_multipath_config() {
@@ -31,8 +31,7 @@ add_res_key_to_multipath_config() {
 
 configure_fence_mpath() {
   add_res_key_to_multipath_config
-  run_command_in_node "${IP_VM01}" "sudo crm configure primitive ${RESOURCE_NAME} stonith:fence_mpath \
-	  params \
+  run_command_in_node "${IP_VM01}" "sudo pcs stonith create ${RESOURCE_NAME} fence_mpath \
 	  pcmk_host_map=\"${VM01}:${NODE1_RES_KEY};${VM02}:${NODE2_RES_KEY};${VM03}:${NODE3_RES_KEY}\" \
 	  pcmk_host_argument=key \
 	  pcmk_monitor_action=metadata \
@@ -46,7 +45,7 @@ test_fence_mpath_is_started() {
   configure_fence_mpath
 
   sleep 15
-  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo crm status")
+  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo pcs status")
   echo "${cluster_status}" | grep "${RESOURCE_NAME}" | grep Started
   assertTrue $?
 }
@@ -66,7 +65,7 @@ test_fence_node_running_the_resource() {
 
   # Check if the node got offline
   sleep 30
-  cluster_status=$(run_command_in_node "${IP_TARGET}" "sudo crm status")
+  cluster_status=$(run_command_in_node "${IP_TARGET}" "sudo pcs status")
   echo "${cluster_status}" | grep "${VM_RESOURCE}" | grep -i offline
   assertTrue $?
 
@@ -75,7 +74,7 @@ test_fence_node_running_the_resource() {
   assertTrue $?
 
   # Check if the resource is still correctly running in another node
-  cluster_status=$(run_command_in_node "${IP_TARGET}" "sudo crm status")
+  cluster_status=$(run_command_in_node "${IP_TARGET}" "sudo pcs status")
   echo "${cluster_status}" | grep "${RESOURCE_NAME}" | grep Started
   assertTrue $?
 }
--- a/ha/virsh/tests/fence_virsh_test.sh
+++ b/ha/virsh/tests/fence_virsh_test.sh
@@ -14,7 +14,7 @@ oneTimeSetUp() {
 }
 
 test_cluster_nodes_are_online() {
-  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo crm status")
+  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo pcs status")
   nodes_online=$(echo "${cluster_status}" | grep -A1 "Node List" | grep Online)
 
   for vm in "${VM01}" "${VM02}" "${VM03}"; do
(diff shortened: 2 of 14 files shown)


ha/virsh/tests/resource_iscsi_target_test.sh
```diff
@@ -41,25 +46,25 @@ test_iscsi_target_resources_are_started() {
...
+  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo pcs status")
   echo "${cluster_status}" | grep "${TARGET_RESOURCE_NAME}" | grep Started
   assertTrue "Could not infer ${TARGET_RESOURCE_NAME} readiness from status output:\n${cluster_status}" $?
   echo "${cluster_status}" | grep "${LUN_RESOURCE_NAME}" | grep Started
   assertTrue "Could not infer ${LUN_RESOURCE_NAME} readiness from status output:\n${cluster_status}" $?
 
   # Check if both resource are running on the same node
   target_location=$(run_command_in_node "${IP_VM01}" \
-			"sudo crm resource locate ${TARGET_RESOURCE_NAME} | cut -d ':' -f2")
+			"sudo pcs status | grep ${TARGET_RESOURCE_NAME} | rev | cut -d ' ' -f1 | rev")
```
panlinux, line 57: That's quite the transformation to get the last column :)

I have one suggestion to make that parsing more obvious, and another to use a more specific `pcs` command. Here is the build-up to the final result:

```
# pcs resource status iscsi-target-rsc 
  * Resource Group: iscsi-target-group:
    * iscsi-target-rsc  (ocf:heartbeat:iSCSITarget):     Started ha-agent-virsh-jammy-resource-iscsi-target-node02
```
Unfortunately another `grep` is needed, so it's not much different than parsing the full output of `pcs status`:
```
# pcs resource status iscsi-target-rsc  | grep iscsi-target-rsc
    * iscsi-target-rsc  (ocf:heartbeat:iSCSITarget):     Started ha-agent-virsh-jammy-resource-iscsi-target-node02
```
And now:
```
# pcs resource status iscsi-target-rsc  | grep iscsi-target-rsc | sed -r 's,.*Started (.*)$,\1,'
ha-agent-virsh-jammy-resource-iscsi-target-node02
```

Another alternative is `awk`:
```
# pcs resource status iscsi-target-rsc  | grep iscsi-target-rsc | awk '{print $5}'
ha-agent-virsh-jammy-resource-iscsi-target-node02
```
In the end, given the lack of output formatting options of `pcs`, and the lack of a `locate` command like `crm`'s, it's a matter of preference. And if you still prefer your double-rev, it's fine. But I kind of prefer to at least gate on the `Started` that we see in the output. That can be added to the initial `grep` perhaps. So you could keep your doble-rev, but add something like `.*Started` to the end of the `grep`.

ha/virsh/tests/fence_virsh_test.sh
```diff
@@ -14,7 +14,7 @@ oneTimeSetUp() {
 }
 
 test_cluster_nodes_are_online() {
-  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo crm status")
+  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo pcs cluster status")
```
panlinux, line 17: In most places you use `pcs status` instead of `pcs cluster status`. I don't mind either way, but it's probably best to be consistent.

lucaskanashiro: The output of `pcs status` and `pcs cluster status` are different as far as I remember. Here I am using `pcs cluster status` because I am checking if the nodes are online. With `pcs status` you get resources information, but yes, with `pcs status` I think you also have information about nodes.

ha/virsh/tests/resource_lvm-activate_test.sh
```diff
@@ -29,22 +29,22 @@ oneTimeSetUp() {
...
 	  vgname=${VOLGROUP} vg_access_mode=system_id"
 }
 
 test_lvm_activate_is_started() {
   configure_cluster_properties
   configure_lvm_activate_resource "${RESOURCE_NAME}" "${VG}"
 
-  run_command_in_node "${IP_VM01}" "sudo crm cluster wait_for_startup"
-  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo crm status")
+  sleep 10
```
panlinux, line 46: `pcs resource create` has a `--wait` option, could that be used in `configure_lvm_activate_resource` when it calls that command? Instead of a `sleep` here?

ha/virsh/tests/resource_lvm-activate_test.sh
```diff
@@ -29,22 +29,22 @@ oneTimeSetUp() {
...
+  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo pcs status")
   echo "${cluster_status}" | grep "${RESOURCE_NAME}" | grep Started
```
panlinux, line 48: This could be one grep only (not something you changed in your branch, I know).

ha/virsh/tests/resource_lvm-activate_test.sh
```diff
@@ -66,9 +66,9 @@ test_move_resource() {
   find_node_to_move_resource "${RESOURCE_NAME}"
 
   # Move resource to another VM
-  run_command_in_node "${IP_VM01}" "sudo crm resource move ${RESOURCE_NAME} ${VM_TARGET}"
+  run_command_in_node "${IP_VM01}" "sudo pcs resource move ${RESOURCE_NAME} ${VM_TARGET}"
   sleep 3
```
panlinux, line 70: Could we also use `--wait` here instead of a hardcoded  `sleep`? It sounds like the `move` command would then block until the operation is finished, or the `wait` period is over, with a proper non-zero exit status if it failed (or hit the wait timeout I suppose).

ha/virsh/tests/resource_systemd_test.sh
```diff
@@ -61,11 +61,11 @@ test_move_resource() {
   find_node_to_move_resource "${RESOURCE_NAME}"
 
   # Move resource to another node
-  run_command_in_node "${IP_VM01}" "sudo crm resource move ${RESOURCE_NAME} ${VM_TARGET}"
+  run_command_in_node "${IP_VM01}" "sudo pcs resource move ${RESOURCE_NAME} ${VM_TARGET}"
   sleep 5
```
panlinux, line 65: Likewise here about `--wait`

ha/virsh/tests/resource_iscsi_target_test.sh
panlinux: At the end of this diff, I see that `get_name_node_running_resource` does the same parsing, but stores the result in a global variable instead of, say echo it to stdout. Was just wondering if we could put this all together in one function. But maybe for another branch if you prefer.

ha/virsh/tests/fence_mpath_test.sh
```diff
@@ -46,7 +45,7 @@ test_fence_mpath_is_started() {
   configure_fence_mpath
 
   sleep 15
-  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo crm status")
+  cluster_status=$(run_command_in_node "${IP_VM01}" "sudo pcs status")
   echo "${cluster_status}" | grep "${RESOURCE_NAME}" | grep Started
```
panlinux, line 49: This could be one `grep` only, but unrelated to your s/crm/pcs/ change.

panlinux (commented):
Ok, done a first pass. Let me know what you think, particularly about the `--wait` comments. Tweaks and clean-ups we can do in follow-up branches.

lucaskanashiro:
Thanks for the review @panlinux . I applied the changes you suggested, only the calls to `grep` I will leave for a next PR. Could you take a look again.

ha/virsh/tests/resource_ipaddr2_test.sh
```diff
@@ -14,23 +14,22 @@ oneTimeSetUp() {
...
 }
 
 configure_ipaddr2_resource() {
   NAME="${1}"
   IP="${2}"
   MASK="${3}"
-  run_command_in_node "${IP_VM01}" "sudo crm configure primitive ${NAME} ocf:heartbeat:IPaddr2 \
-	  ip=${IP} cidr_netmask=${MASK} op monitor interval=30s"
+  run_command_in_node "${IP_VM01}" "sudo pcs resource create ${NAME} ocf:heartbeat:IPaddr2 \
+	  ip=${IP} cidr_netmask=${MASK} op monitor interval=30s --wait"
```
panlinux, line 25: I was checking the `pcs` manpage again, and `--wait` without a value defaults to 1h. Maybe we should set a value closer to the `sleep` values you had before? Perhaps 60s, to give it some margin? Not just here, but in all `--wait` calls

panlinux:
Tests passed now, nice. +1 from me

lucaskanashiro:
@panlinux could you make a final check before we merge it?

panlinux:
+1
(I don't see an "approve" button/link)

panlinux (approved):
Ah, found it

+1

lucaskanashiro:
Thanks, I am merging it.

PR #168, opened 2022-11-14
cleanup: update lxd prefix for pro client behave

orndorffgrant:
As part of the rename of UA to pro, we're changing the prefix of behave lxd instances for the pro client integration tests.

This is happening here: https://github.com/canonical/ubuntu-advantage-client/pull/2303

I'm not sure where this script is called. Because the prefix can be overridden, we should make sure any place we call this script is calling it correctly as well.


paride (approved):
LGTM, thanks!

PR #172, opened 2023-03-20
metric: add more disk and package info

cpaelzer:
This is needed as a lessons learned from checking the results.


paride (approved):
Looks good!

cpaelzer:
Thanks, turns out the permissions on the test servers when run with -e make it break.
I'll have a hot fix shortly, if trivial I'll land it right away - if not expect a PR soon.

PR #173, opened 2023-03-24
metric: Run most early measurements before adding dependencies

cpaelzer:
Most early stage measurements can run before installing dependencies like hyperfine. Therefore move them to not be seen e.g. in disk or package statistics.


cpaelzer:
This is one more little thing I found we can optimize, if you agree merge it and I'll put an annotation to the changed results after the weekend.

paride (approved):
Looks good, thanks!

PR #174, opened 2023-04-06
gcp: delete active instances older than two weeks

lucasmoura:
Since we had a problem with the cleanup of old gcp instances, we are now cleaning up any active instance that is running for more than two weeks

### DIFF
--- a/ubuntu-advantage-client/gcp_cleanup.py
+++ b/ubuntu-advantage-client/gcp_cleanup.py
@@ -70,7 +70,17 @@ def clean_gcp(credentials_path, project_id, tag, before_date, region, zone):
             instance["creationTimestamp"].split("T")[0], "%Y-%M-%d"
         )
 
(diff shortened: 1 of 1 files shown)


renanrodrigo:
I am not sure about the cloud-init side, but I'd say one week is enough; unless they expect some test to take many days of course

TheRealFalcon (approved):
On cloud-init, the time to wait could be much lower. We never have jobs running more than a few hours, so even a couple of days would be fine for us.

lucasmoura:
Got it, I will update it to 2 days then

renanrodrigo (approved):

paride (approved):
LGTM, but I need to check again where/when we are running this in Jenkins. I'll circle back here on Tuesday (I'm off next Monday 2023-04-10).

paride:
I still plan to review / improve how we are cleaning stuff after integration testing runs, but in any case this looks good to land.

PR #175, opened 2023-04-18
oci: add support for libmemcached >= 1.1.0

athos-ribeiro:
Before libmemcached 1.1.0 the  memccat CLI would not crash in case a key was not provided, but would crash if an empty key was passed. This behavior was changed to crashing when not providing a key and not crashing when an empty key is provided since 1.1.0. Therefore, we test both cases to ensure this test is compliant to all supported libmemcached versions.


paride (approved):
LGTM

PR #176, opened 2023-04-18
oci: force volume removal for squid tests

athos-ribeiro:
This may cause a test failure in any situation where this volume cannot be removed. Since we are using the volume name, which aims this specific test, it should be safe to go for a forced removal.


paride (approved):
lgtm, thanks!

athos-ribeiro:
Thanks, Paride!

PR #179, opened 2023-06-23
Add metric server bugs

cpaelzer:
This is a tool from @mclemenceau that I extended to put data into google and here changed it to work with anonymous user and to be in the script directory used for our automation jobs.

### DIFF
--- a/metric-server-bugs/lp-query-gspread
+++ b/metric-server-bugs/lp-query-gspread
@@ -32,7 +32,7 @@ parser.add_argument("--subscriber", action='store_true',
                     help="do not report all, but only subscribed bugs")
 args = parser.parse_args()
 
-launchpad = Launchpad.login_with(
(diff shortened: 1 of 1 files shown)


cpaelzer:
Currently this works, up to the point when it gets stuck authenticating with gspread.
I think to fully make it work we will need to request [this](https://support.google.com/a/answer/2589954?hl=en) in the firewall rules.

cpaelzer:
We passed the firewall issues, since this was all http[s] proxies were enough.
I'll ping once it fully worked to then review and merge.

cpaelzer:
And with https://jenkins.canonical.com/server-team/view/metrics/job/metric-server-bugs/9/ we have the first good run.
This would be ready for review and merge.

paride (approved):
LGTM

PR #180, opened 2023-06-30
cloud-init/subiquity: use bsdtar and avoid kvm -daemonize for jenkins runner

blackboxsw:
- Adapt from mounts to bsdtar  to extract files from ISO.

- Also avoid kvm -daemonize as Jenkins runners won't have access to the display.
  Wrap launch_kvm for first boot in a background thread which writes console log to first-boot-console.log instead of saturating 
  stdout with all console boot messages and shutdown messages for system first boot.

### DIFF
--- a/cloud-init/subiquity_live_server.py
+++ b/cloud-init/subiquity_live_server.py
@@ -22,6 +22,7 @@ https://github.com/canonical/pycloudlib.
 """
 
 import argparse
+from functools import partial
 import json
 import logging
 import os
@@ -31,6 +32,7 @@ import socket
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 from enum import Enum
 from functools import wraps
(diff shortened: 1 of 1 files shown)


paride (changes requested):
There is also an `umount` command that needs root permissions. However I fear that we're going to leak mounts in one way of another, which I'd like to avoid. In general I think it's better to avoid sudo commands in Jenkins.

Looks like you only need to extract `vmlinuz` and `initrd` from the ISO, is that correct? One way to do it without mounting is via `bsdtar`, e.g.

```
bsdtar -v -x -f ubuntu-22.04.2-desktop-amd64.iso casper/vmlinuz
```
This will extract `casper/vmlinuz` as `casper/vmlinuz` locally. To extract it without path you could use `--strip-components 1` or `-O` and then redirect stdout.

Would this work in the context of this script?

blackboxsw:
> There is also an `umount` command that needs root permissions. However I fear that we're going to leak mounts in one way of another, which I'd like to avoid. In general I think it's better to avoid sudo commands in Jenkins.
> 
> Looks like you only need to extract `vmlinuz` and `initrd` from the ISO, is that correct? One way to do it without mounting is via `bsdtar`, e.g.
> 
> ```
> bsdtar -x -f ubuntu-22.04.2-desktop-amd64.iso casper/vmlinuz
> ```
> 
> This will extract `casper/vmlinuz` as `casper/vmlinuz` locally. To extract it without path you could use `--strip-components 1` or `-O` and then redirect stdout.
> 
> Would this work in the context of this script?

@paride excellent suggestion yes. bsdtar will help here. and looks to work well for anything that mattered in this script.

blackboxsw:
Note: [Jenkins test run showing success on script run](https://jenkins.canonical.com/server-team/job/zzz-bbsw-subiquity-test/15/console). I still need to work on a real pipeline that grabs correct log artifacts.

paride (approved):
Looks good!

PR #181, opened 2023-07-07
wip: Add metric pro stats

cpaelzer:


metric-pro-stats/auto-update-stats
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -uxe
+# This meant to run each Sun->Mon night checking what it might need to update
+[[ "$(LC_ALL=C date +%A)" == "Monday" ]] || exit 1
+echo "Sanity check passed, this is a Monday"
+
+path="$(dirname "$0")"
+defaultoptions="--log info --cachefile /tmp/esm-stats-new.cache --append --connectionenv ~/work/credentials/esm-stats/ps5-prod-esm-cache-ro.novarc ~/work/credentials/esm-stats/ps4.5-prod-esm-cache-ro.novarc"
```
paride, line 8: Let's not hardcode /tmp and give precedence to $TMPDIR if set (so we'll use a job specific tmpdir in Jenkins). Also that `~work/` path won't work in Jenkins, but I guess that's part of the WIP state.

metric-pro-stats/auto-update-stats
```diff
@@ -0,0 +1,36 @@
...
+[[ $(date +%d) -gt 7 ]] || exit
+echo "Sanity check passed, this is the first Monday of a month"
+
+# shellcheck disable=SC2016
+mend="$(date -d '$(date +%Y/%m/01) - 1 day 00:00' +'%Y-%m-01')"
+# shellcheck disable=SC2016
+mstart="$(date -d '$(date +%Y/%m/01) - 1 day 00:00' +'%Y-%m-%d')"
+echo "Monthly run - collecting last month from ${mstart} to ${mend}"
+# shellcheck disable=SC2086
+"${path}/server-pro-statistics.py" \
```
paride, line 31: Is this one missing? Is it actually going to consume `$mstart` and `$mend` with the that verbatim `$( ... )` expansion?

paride:
I didn't do an in-depth review, but the general approach LGTM and will integrate well with the other tools/metrics we have. I added a couple of inline comments, but again this is not a full review.

metric-pro-stats/auto-update-stats
cpaelzer: Now I feel bad `auto-update-stats` is only an artifact of some former local experiments. That is now properly split and much better in the yaml. I'll remove it - sorry for the noise.

metric-pro-stats/auto-update-stats
cpaelzer: Now I feel bad `auto-update-stats` is only an artifact of some former local experiments. That is now properly split and much better in the yaml. I'll remove it - sorry for the noise.

cpaelzer:
Since I'm still waiting on the api access to the ppa data I'll convert this to a draft for now.

PR #183, opened 2023-07-10
metric-server-simple: deepen disk report

cpaelzer:
We quite often see fluctuations in /usr/lib or /var/cache and going one level deeper would help a lot to point to the issue more quickly e.g. /var/cache/snap.


aciba90 (approved):
+1, thanks!

PR #184, opened 2023-07-10
metric-server-simple: fix service-security

cpaelzer:
LXD by default does not spawn a login shell, but that is required to get a systemd user scope session.

Without a login shell the context is missing and this will fail with `Failed to connect to bus: No medium found`


cpaelzer:
Issue is seen e.g. in https://jenkins.canonical.com/server-team/view/metrics/job/metric-server-simple-lunar-vm-c1-m1/125/console

Example with the fix from a local run:

```
$ lxc exec metric-server-simple-lunar-vm-c1-m1 --env=DEBIAN_FRONTEND=noninteractive -- su ubuntu --login -c 'systemd-analyze security --no-pager --user'
UNIT                        EXPOSURE PREDICATE HAPPY
...
```

aciba90 (approved):
+1, thanks!

PR #185, opened 2023-07-17
metric-server-simple: gather service status

cpaelzer:
Sometimes a rare event might happen that is hard to recerate and debug later. So far we have seen two times that service startup was bad.

Therefore it would be very helpful to have at least the general final status of all services in the log as it might indicate what was broken.


aciba90 (approved):
LGTM, thanks!

paride (approved):
LGTM

Issue #188, opened 2023-08-03
server-simple: service status logged for host system and not for system under test


paride:
@cpaelzer looks like we have issues here:

https://github.com/canonical/server-test-scripts/blob/f444e6e9603d490564a306401fd1e693c6eb59aa/metric-server-simple/metric-server-simple.sh#L194

1. We're executing it on the host system (Jenkins node) and not in the container, right? We want a `cexec` there I believe. Let's also add `--no-pager`.
2. The command can return nonzero, so we need a `|| true`.
3. Can we save that to file instead of printing to stdout? The output is very long and makes the Jenkins consolelog even more difficult to read. If you are +1 on saving to file we'll also need to archive it, adding it to the `archiveArtifacts('results-*.json')` line in the JJB definitions

cpaelzer:
yeah, we came to the same conclusion.
I had all but Cexec ready, adding that to #190

PR #190, opened 2023-08-04
metric-server-simple: ignore RC of combined service status

cpaelzer:
The RC of systemctl status --all can easily be non-zero while not being fatal for the system or the metric run. Ignore that to not run into retries (wasting time) and then giving up eventually.

Fixes: 7f331de8


paride:
@cpaelzer I think we're also missing a cexec, see see https://github.com/canonical/server-test-scripts/issues/188

cpaelzer:
> @cpaelzer I think we're also missing a cexec, see see #188

Ack, but IMHO we should use Cexec here as sometimes output is reduced if not root.

paride (approved):

PR #191, opened 2023-08-04
metric-server-simple: fix status redirect

cpaelzer:
The || true recently added did break the redirection which no more is capturing the output of systemd-analyze. Since `true` has no output the file isn't created at all and breaks the import.

Change the order so that the output of systemd-analyze is redirected.

Fixes: f444e6e9


paride (approved):

paride:
(pedantic) The issue is not `true` not having output (`true > foo` creates `foo` just fine), but it's that in my buggy code:
```
Cexec su ubuntu --login -c 'systemd-analyze security --no-pager --user' || true > "${resultfile}"
```
the redirection only happens if the right hand of the `||` is executed. So basically it was only creating an (empty, useless) log file if `systemd-analyze` failed, otherwise no file creation. Very buggy!

PR #192, opened 2023-08-07
metric-server-simple: do not trace flaky cache values

cpaelzer:
/var/cache/apt can fluctuate a lot depending on how many packages might need to be updated as (relative to the daily image) to install packages of the workload. We are not interested in those fluctuations as they would auto-clean over time anyway.

Therefore run `apt clean` after installing packages for the loaded stage, but we want to know the initial image size if e.g. a clean was forgotten on build, so do not clean it there.

### DIFF
--- a/metric-server-simple/data2influx.py
+++ b/metric-server-simple/data2influx.py
@@ -125,7 +125,7 @@ def parse_cpustat_measurement(fname, point):
 
 
 def parse_disk_measurement(fname, point):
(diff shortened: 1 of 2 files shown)


cpaelzer:
FYI - belongs to https://github.com/canonical/server-jenkins-jobs/pull/285

paride (approved):
LGTM, thanks!

PR #194, opened 2023-09-27
Add openssh-ccache-check script and README

panlinux:
This will be called by a jenkins job.

### DIFF
--- a/openssh-ccache-check/openssh-ccache-check.py
+++ b/openssh-ccache-check/openssh-ccache-check.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+
+import sys
+
+from subprocess import check_output, call
+try:
+    from launchpadlib.launchpad import Launchpad
+except ImportError:
+    print("Failed to import launchpadlib. Please install python3-launchpadlib")
+    sys.exit(1)
+
+
+class RmadisonPackage(object):
+
+    def __init__(self, package):
+        self.package = package
+        self.rmadison = self._get_rmadison()
+
+    def _get_rmadison(self):
+        cmd = ["rmadison", "-asource", self.package]
+        return self._parse_rmadison(check_output(cmd))
+
+    def _parse_rmadison(self, data):
+        pocket_version = {}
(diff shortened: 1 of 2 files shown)


openssh-ccache-check/openssh-ccache-check.py
```diff
@@ -0,0 +1,114 @@
...
+            warn = True
+        print()
+    if warn:
+        print("ACTION NEEDED")
+    else:
+        print("ALL GOOD")
+    return warn
+
+if __name__ == "__main__":
+    sys.exit(main())
```
paride, line 114: I found it slightly counterintuitive to rely on `bool(False) == 0` to exit with RC=0, but I see no actual issues with that.

paride (changes requested):
LGTM, but there are some minor style issues, for example flake8 outputs:

```
openssh-ccache-check/openssh-ccache-check.py:42:31: E231 missing whitespace after ','
openssh-ccache-check/openssh-ccache-check.py:45:13: E125 continuation line with same indent as next logical line
openssh-ccache-check/openssh-ccache-check.py:46:80: E501 line too long (85 > 79 characters)
openssh-ccache-check/openssh-ccache-check.py:50:1: W293 blank line contains whitespace
openssh-ccache-check/openssh-ccache-check.py:53:80: E501 line too long (83 > 79 characters)
openssh-ccache-check/openssh-ccache-check.py:98:15: E231 missing whitespace after ','
openssh-ccache-check/openssh-ccache-check.py:113:1: E305 expected 2 blank lines after class or function definition, found 1
```
These can all be auto-fixed by running
```
black -l 80 openssh-ccache-check.py
```
Given that getting it nicely formatted and linted is that easy, let's go for it.

(Ideally I should add some linting checks in CI for this repo, including black.)

openssh-ccache-check/openssh-ccache-check.py
panlinux: Yeah, that makes sense. Changing.

PR #199, opened 2023-12-07
ha: do not assume all homes are /home/$USER

athos-ribeiro:
When setting up the cluster, we have been assuming that all user home directories are in /home/$USER. That is not necessarily true and it is not true for the default root user home which is in /root.


lucaskanashiro (approved):
I do not think this is a real issue right now because we do not run tests as root AFAIK, but it is indeed an improvement.

PR #204, opened 2024-01-11
ha: add initial pgsql resource agent test

athos-ribeiro:
This tests a simple master/slave replication setup as described in the upstream documentation in [1].

First, it sets up an PostgreSQL HA cluster with 2 nodes in a master/slave configuration. Then, it simulates a failure in the master node to verify if the slave is successfully promoted.

This does not use a floating IP address for service access since it would depend on another resource agent.

While shunit2 is a xUnit based framework, it does execute the tests in the order they are defined by default. This test script does rely on the order each test is called (i.e., the tests are not independent). This was done this way to avoid resetting the cluster state for each test we want to perform since this would require more resources for our runs.

[1] https://wiki.clusterlabs.org/wiki/PgSQL_Replicated_Cluster

### DIFF
--- a/ha/virsh/tests/resource_pgsql_test.sh
+++ b/ha/virsh/tests/resource_pgsql_test.sh
@@ -1,0 +1,175 @@
+# shellcheck shell=bash
+
+# Based on LP: #2013084; and
+# https://wiki.clusterlabs.org/wiki/PgSQL_Replicated_Cluster; and
+# https://clusterlabs.github.io/PAF
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+# shellcheck disable=SC1090
+. "$(dirname "$0")/../vm_utils.sh"
+
+
+setup_pgsql() {
+  PG_HBA="\
+host    all             all     127.0.0.1/32        trust\n\
+host    all             all     192.168.0.0/16      trust\n\
+host    replication     all     192.168.0.0/16      trust"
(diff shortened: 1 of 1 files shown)


ha/virsh/tests/resource_pgsql_test.sh
```diff
@@ -0,0 +1,175 @@
...
+# https://wiki.clusterlabs.org/wiki/PgSQL_Replicated_Cluster; and
+# https://clusterlabs.github.io/PAF
+
+# shellcheck disable=SC1090
+. "$(dirname "$0")/helper/test_helper.sh"
+# shellcheck disable=SC1090
+. "$(dirname "$0")/../vm_utils.sh"
+
+
+setup_pgsql() {
```
lucaskanashiro, line 13: I'd split this function into multiple ones to make it easier to understand, such as one to set up the cluster properly, another one to set up the master and another one to the replica.

lucaskanashiro (approved):
I added a simple suggestion but apart from that LGTM.

ha/virsh/tests/resource_pgsql_test.sh
athos-ribeiro: Thanks, Lucas! I split the functions where it made sense without changing the order of the calls in that single (big) function.

lucaskanashiro:
Thanks @athos-ribeiro ! LGTM.

lucaskanashiro (approved):

PR #206, opened 2024-01-29
m/metric-server-simple.sh: Improve handling of "cloud-init status --wait"

sergiodj:
From Noble onwards, "cloud-init status --wait" can return other than
zero.  The script should only error out if the return code is 1.

@paride, this fixes a problem which has been affecting the Noble
metrics VM for a several days now.  WDYT?

Thanks!


cpaelzer:
+1, that works

PR #210, opened 2024-02-05
ha: fix flakiness on ipmilan tests

athos-ribeiro:


ha/virsh/tests/fence_ipmilan_test.sh
```diff
@@ -52,8 +52,11 @@ get_tester_ip() {
 }
 
 setup_tester() {
-  run_command_in_node "${IP_TESTER}" "sudo apt-get update && \
-    sudo apt-get install -y ipmitool fence-agents-base"
+  APT_OUTPUT=$(run_command_in_node "${IP_TESTER}" "DEBIAN_FRONTEND=noninteractive sudo apt-get update && sudo apt-get install -y ipmitool fence-agents-base")
```
paride, line 55: Some comments/questions:
- Why are we saving output to `APT_OUTPUT`? I don't see that being used.
- If we actually need that, let's pass `-q` or `-qq` to apt-get. One `-q` is what the manpage recommends for logging.
- `DEBIAN_FRONTEND=noninteractive` won't work:
  - sudo will remove it for the environment, we need the `sudo VAR=value command` syntax
  - in any case, it would be valid only for the left side of the `&&`.

ha/virsh/tests/fence_ipmilan_test.sh
```diff
@@ -52,8 +52,11 @@ get_tester_ip() {
...
+  if [ $? -ne 0 ]; then
```
paride, line 56: Is this `if` needed because we want to explicitly return 1 when `$?` is nonzero for some reason?

paride (changes requested):

ha/virsh/tests/fence_ipmilan_test.sh
athos-ribeiro: Thanks @paride!

>     * Why are we saving output to `APT_OUTPUT`? I don't see that being used.

We actually do not need that output. I just wanted to run the command in a child shell since `run_command_in_node` exits on failures. Recently, I have been seeing failures due to lock-acquiring issues with dpkg (as @lucaskanashiro mentioned he was seeing as well - this time, it also happened in jenkins).  Here, backing-off seems to be a reasonable workaround until I find the root cause for the issue.

If the command output is not captured nor discarded, it will be passed to the parent shell as regular commands. I'd be happy to adopt any other suggestions here.

>     * `DEBIAN_FRONTEND=noninteractive` won't work:
>       
>       * sudo will remove it for the environment, we need the `sudo VAR=value command` syntax
>       * in any case, it would be valid only for the left side of the `&&`.

Nice catch! Fixed.

paride: To run in a subshell maybe enclosing the command within `( )` will suffice?
```
#!/bin/bash
echo hi
(echo hi from subshell)
```

ha/virsh/tests/fence_ipmilan_test.sh
paride: I see, but I the function should return the RC of the latest command:
```
$ foo() { (exit 42) }
$ foo
$ echo $?
42
```
so I'm still not sure of why that `if` is needed...

athos-ribeiro: I was changing the error status code to 1.

Let's preserve it then. Fixed

paride (approved):
Thanks!

PR #211, opened 2024-02-26
ha: use dhcpcd instead of dhclient in noble

athos-ribeiro:
dhcpcd should be the new supported dhcp client starting from noble.

Here is the failure we started seeing in all noble related jobs:

https://jenkins.canonical.com/server-team/view/ha/job/ha-agent-virsh-noble-resource_pgsql/33/console

And here is a successful run based on the proposed branch:

https://jenkins.canonical.com/server-team/view/ha/job/ha-agent-virsh-noble-resource_pgsql/35/console


ha/virsh/vm_utils.sh
```diff
@@ -22,6 +22,11 @@ CLUSTER_NAME="test-cluster"
 HA_NETWORK="ha-${UBUNTU_SERIES::1}-${AGENT}"
 HA_NETWORK=${HA_NETWORK::15}
 
+DHCP_CLIENT="dhclient"
+if [[ "${UBUNTU_SERIES}" = "noble" ]]; then
```
paride, line 26: This however will break post-noble, right? Should we do the opposite instead? Something like this (untested!)


```
DHCP_CLIENT=dhcpcd
if [[ "${UBUNTU_SERIES}" = "jammy" ]]; then # and maybe mantic? or let's remove it?
    DHCP_CLIENT="dhclient"
fi
```

paride (commented):
See inline question.

ha/virsh/vm_utils.sh
athos-ribeiro: Thanks Paride! +1; I will adjust this (and another place where we are checking for noble) and update this PR.

sergiodj: Hey,

When using `[[` (bash-only), the string comparison operator should be `==`.  I think there is some lenience in accepting `=`, but IMHO it's better to stick with `==` here.

Just thought I'd give my 1/2 cent ;).

athos-ribeiro:
Thanks for the reviews!

I addressed your comments and force pushed it here!

sergiodj:
+1 from me, thanks Athos!

athos-ribeiro:
Thanks!

PR #212, opened 2024-06-04
openssh ccache: add noble

panlinux:
This adds noble to the openssh-ccache-check script used by the openssh-ccache-check jenkins job.


paride (approved):
LGTM

